PR #5, opened 2021-05-12
add data cookiecutter

elpamart:
The PR contains the data cookiecutter in a self-contained data folder so we avoid having python on the root.


tiagojsag:
not sure if the github stuff makes sense, as this is a subfolder, but otherwise LGTM

data/.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+
+[*]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+insert_final_newline = true
+charset = utf-8
+end_of_line = lf
+
+[*.bat]
```
tiagojsag, line 13: .bat files?

image

data/test_environment.py
```diff
@@ -0,0 +1,25 @@
+import sys
+
+REQUIRED_PYTHON = "python3"
+
+
+def main():
+    system_major = sys.version_info.major
+    if REQUIRED_PYTHON == "python":
+        required_major = 2
```
tiagojsag, line 9: python2 is dead, please put it out of its misery

tiagojsag:
- I think the Github templates dont really make sense in a subfolder, we should probably agree on a common format for all PRs - happy to use this one
- Overall, I see a lot of boilerplate code, dependencies and such. I am assuming this is some sort of copy-paste from a template, and these are all things you dont need now but may need in the future, in which case I would suggest removing pretty much all of them, and re-add them as they become needed.  This ensures you'll only add the dependencies you'll actually need, as opposed to starting with everything and never really cleaning up, because at some point it becomes too complicated to do that cleaning.

data/notebooks/requirements.txt
```diff
@@ -0,0 +1,40 @@
+
+
+# configuration requirements
+
+python-dotenv>=0.5.1
+
+# external tools requirements
+tqdm
```
tiagojsag, line 8: every time someone adds a dependency on an external package without specifying a version, god kills a puppy :P

In all seriousness, this is a recipe for disaster. If one of these packages releases a major version in the future, with a full refactor of their API, your whole application will stop working, and you'll have no control over it.

elpamart: I've already added the versions to all the packages

data/test_environment.py
elpamart: removed!

elpamart:
@tiagojsag I've added an issue on the main cookicutter to simplify it once we start testing it here and in marxan. The idea is to have some boilerplatte that makes our lives easier but we would like to identify those parts that are not needed and therefore can be removed. 

can we merge it as it is right now?

elpamart:
@tiagojsag  I've update the versions again. Please merge if you are happy with this (or I can do it if you say so)

tiagojsag:
LGTM

PR #10, opened 2021-05-14
chore:: LANDGRIFFON-50 create s3 bucket

alexeh:


infrastructure/variables.tf
```diff
@@ -31,9 +31,16 @@ variable "tf_state_bucket" {
   description = "The name of the S3 bucket where the state is stored"
 }
 
+variable "raw_data_bucket" {
+  type        = string
+  default     = "landgriffon-raw-data"
+  description = "The name of the s3 buckets where where raw data will be stored"
```
tiagojsag, line 37: review the description please

tiagojsag (changes requested):
- Not really sure why .terraform.lock.hcl was removed. It's my understanding it's a best practice to keep it under VCS but I may be wrong
- Your module defines a bunch of variables, but then you never use them. I suspect you need to do a bit more reading on the subject, and understand them better.

infrastructure/variables.tf
alexeh: Fixed [here](https://github.com/Vizzuality/landgriffon/pull/10/commits/714a6d7cc7974b7aa38f326f7dfd6af112c8c399)

alexeh:
> * Not really sure why .terraform.lock.hcl was removed. It's my understanding it's a best practice to keep it under VCS but I may be wrong

-I was not sure about that, the side-effects that could produce, but actually makes sense having it on local. Thanks for pointing out

> * Your module defines a bunch of variables, but then you never use them. I suspect you need to do a bit more reading on the subject, and understand them better.

Changes [here](https://github.com/Vizzuality/landgriffon/pull/10/commits/83085fa97a4498198c1c99d0ed4b41b79cb195d0)

And reading about TF modules as well 👍

alexeh:
@tiagojsag changes applied 6c248b0e9fc31d79b13c46d0a9905ccdedadff45

PR #14, opened 2021-05-20
add h3 data processing test

elpamart:
h3 processing for polygon area


elpamart:
@01painadam @fgassert I've more or less fixed the issue by finding the nearest point with value with shapely. I'm sure there would be a more efficient way of doing this but we can go with this for now.

01painadam (approved):
A great demonstration of the capabilities (and challenges) of H3.

PR #15, opened 2021-05-31
add notebook for water indicator prototype

elpamart:
Add notebook with water indicator. This notebook will need to be updated to cover the area approach we have highlighted in this doc: https://docs.google.com/document/d/1GVUEv8l9T3zogKvIThCUAWN_tbZdpu2-MbE3PyOQepw/edit#heading=h.pt3naz996cor


01painadam:
So some general functions we could extract to make this a bit more repeatable (not for now):

1. Extract info (output things like pixel size, corner coordinates using gdal info)
2. Rasterise (tages target raster, rasterises to its resolution/extent)
3. Raster math (lots of calculations where we alias a target path against a variable - this could be nice to have a `variable: path` mapping as an input, along with the algebraic equation)
4. Hex aggregation math
5. Hex to gdf

Otherwise, looks great! Nice one with the Assignation Function (f') demo 💪

01painadam (approved):

PR #17, opened 2021-06-01
add first steps in biodiversity metric calc

elpamart:
First steps in the calculation of the biodiversity indicator. For that, the notebook imports the regional taxa aggregated characterization factor from the LCIA recommendations at ecoregion level. These data is rasterizaed with the same resolution of the area distributed one and will need to be multiplied by the deforestation for each commodity.

Note: the characterization raster just covers the median of PDF/m2 in annual crops, we may want to consider other scenarios but for the time being we can simplify with this use case.


01painadam (approved):
Great starting point - we can complete the prototype once we have the deforestation risk layer

PR #21, opened 2021-06-07
chore: LANDGRIFFON-45 - Configure CI/CD for Frontend

alexeh:
This PR adds 3 changes:

-A YML file with the CI conf for GH Actions. Right now there are only Cypress tests, and the default ones, so I named the job jus as 'Frontend tests'

-A couple of options on `cypress.json` just to avoid generating artifacts (screenshots, videos) if tests success, since I believe it doesnt make much sense to check these reports if all went fine. We can add a exclusive CI cypress conf for CI if you still want to generate said artifacts locally on test success

-A script in clients `package.json` just for CI, which will load the app in dev mode and run cypress in headless mode


.github/workflows/frontend-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+  NODE_ENV: test
+
+defaults:
+  run:
+    working-directory: client
+
+jobs:
+  frontend-tests:
+    name: Frontend Tests
+    runs-on: ubuntu-latest
```
tiagojsag, line 20: please pin a version, as "latest" will change every 6 months

tiagojsag (changes requested):
one minor comment on the code already present.
One thing that seems to be missing altogether is code coverage for the frontend.

.github/workflows/frontend-tests.yml
alexeh: @tiagojsag good catch. Changes 417e5d9e207e7bdf039fcc01a26b3824a150f740

PR #22, opened 2021-06-09
add modification in purchased area probability methodology

elpamart:
slightly modification of the probability area calculations.


elpamart:
@01painadam I've added the modification of the probability purchase area, the calculation of the biodiversity risk map and metric and the deforestation metric. Please let me know what do you think.

01painadam:
Great! The area probability is much easier to follow - nice one :)

01painadam (approved):

PR #24, opened 2021-06-10
chore: LANDGRIF-45 - Configure Docker for Frontend

alexeh:
This PR adds:

[`Dockerfile`](https://github.com/Vizzuality/landgriffon/compare/chore/LANDGRIF-43-configure-docker-frontend?expand=1#diff-c0a19429aa475a5c4fba2e72499771095a4d55a72adf8271238b2a402928b475) file for `client` application

Update [`docker-compose.yml`](https://github.com/Vizzuality/landgriffon/compare/chore/LANDGRIF-43-configure-docker-frontend?expand=1#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3) adding client service

Update [`Makefile`](https://github.com/Vizzuality/landgriffon/compare/chore/LANDGRIF-43-configure-docker-frontend?expand=1#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52) to start just the frontend app or all services at once

New [`ENV VAR`](https://github.com/Vizzuality/landgriffon/compare/chore/LANDGRIF-43-configure-docker-frontend?expand=1#diff-14d0b30702e03f0c4431c6bb68320bc3b0d1660ab34d9e501d28b503d424d42a) **REQUIRED** for exposing the client outside the docker network

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,13 @@ test-commands:
 start-api:
 	docker-compose --project-name ${COMPOSE_PROJECT_NAME} up --build api
 
+# Starts the CLIENT application
+start-client:
+	docker-compose --project-name ${COMPOSE_PROJECT_NAME} up --build client
+
 # Start all the services.
 start:
-	docker-compose --project-name ${COMPOSE_PROJECT_NAME} up --build api
+	docker-compose --project-name ${COMPOSE_PROJECT_NAME} up --build
 
 stop:
 	docker-compose $(DOCKER_COMPOSE_FILE) stop
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -1,0 +1,32 @@
+FROM node:14.17-alpine3.13
+LABEL maintainer="hello@vizzuality.com"
+
+ARG UID=5000
+ARG GID=5000
+
+ENV NAME landgriffon-client
+ENV USER $NAME
+ENV APP_HOME /opt/$NAME
+
+RUN addgroup -g $GID $USER &&  adduser -u $UID -D -G $USER $USER
+
+WORKDIR $APP_HOME
+RUN chown $USER:$USER $APP_HOME
+
+USER $USER
+
+COPY --chown=$USER:$USER package.json yarn.lock ./
+RUN yarn install --frozen-lockfile
+
+# NextJS project folders
+COPY --chown=$USER:$USER components containers lib pages public services styles svgs ./
+# NextJS config files
+COPY --chown=$USER:$USER next.config.js next-env.d.ts now.json postcss.config.js tailwind.config.js ./
+# Testing folders / files
+COPY --chown=$USER:$USER cypress cypress.json ./
+# Project config giles
+COPY --chown=$USER:$USER .* ./
+COPY --chown=$USER:$USER entrypoint.sh tsconfig.json ./
+
+EXPOSE 3000
+ENTRYPOINT ["./entrypoint.sh"]
--- a/client/entrypoint.sh
+++ b/client/entrypoint.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+set -e
+
+case "$1" in
+    develop)
+        echo "Running web application in development mode"
+        exec yarn dev
+        ;;
+    test)
+        echo "Running Tests"
+        exec yarn cypress:run
+        ;;
+    build)
+        echo "Build web application"
+        exec yarn build
+        ;;
+    start)
+        echo "Run previously built application"
+        exec yarn start
+        ;;
+    *)
+        echo "Usage: entrypoint.sh {develop|test|build|start}" >&2
+        exit 1
+        ;;
+esac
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -31,9 +31,13 @@ test-commands:
...
 	docker-compose --project-name ${COMPOSE_PROJECT_NAME} up --build api
 
+# Starts the CLIENT application
+start-client:
+	docker-compose --project-name ${COMPOSE_PROJECT_NAME} up --build client
+
 # Start all the services.
 start:
-	docker-compose --project-name ${COMPOSE_PROJECT_NAME} up --build api
+	docker-compose --project-name ${COMPOSE_PROJECT_NAME} up --build
```
alexeh, line 40: Not totally sure if we want to force to build images on start every time. It may make sense or not.
Let me know your thoughts @tiagojsag

client/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+COPY --chown=$USER:$USER package.json yarn.lock ./
+RUN yarn install --frozen-lockfile
+
+# NextJS project folders
+COPY --chown=$USER:$USER .next .storybook components containers cypress docs lib pages public services styles svgs ./
+# NextJS config files
+COPY --chown=$USER:$USER next.config.js next-env.d.ts now.json postcss.config.js tailwind.config.js ./
+# Project config giles
+COPY --chown=$USER:$USER .* ./
```
alexeh, line 26: I believe we can avoid passing linting and development related config stuff to the container, but wasn't sure either.

Makefile
tiagojsag: depends on whether docker picks up changes automatically or not.

client/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM node:14.16.1-alpine3.13
```
tiagojsag, line 1: FE uses 14.17

client/Dockerfile
tiagojsag: you can, assuming you don't want to run the tests inside the container (we I think we don't need to). @davidsingal thoughts on this? Are we missing anything needed for the build process?

client/entrypoint.sh
```diff
@@ -0,0 +1,21 @@
...
+    test)
+        echo "Running Tests"
+        exec yarn cypress:run
+        ;;
+    start)
+        echo "Running web application in production mode"
+        exec yarn start:prod
+        ;;
+    *)
+        echo "Usage: entrypoint.sh {develop|test|start:prod}" >&2
```
tiagojsag, line 18: start:prod -> start

client/Dockerfile
tiagojsag: the BE image has the test stuff, so maybe add it here too...?

alexeh: I'm adding all the testing stuff so it can run inside the container (just in case) but I was referring mostly to stuff like eslint conf, which (i think) shouldn't be necessary inside docker ecosystem

client/entrypoint.sh
alexeh: `start` needs a previous build to run prod mode, which is not present when container runs
Thats why I added a [new script](https://github.com/Vizzuality/landgriffon/pull/24/files#diff-1846122c2c83a486a3693f7966aa522c34cf489f674185c4da0d9221683fd81f) in `/client/package.json`

Makefile
alexeh: It does in dev mode

client/Dockerfile
alexeh: @tiagojsag nice catch. -> 2e0e070956334f4b76ac05cf30317bac937be5ab

client/Dockerfile
alexeh: I was adding cypress folder as part of NextJS folders copy and I was forgetting the config file.
Thanks for pointing out. Changes -> b17c4bfc1e158409c45467696150de180caaa457

davidsingal: I don't think we need run tests inside the container. Will it help you if I move everything inside a `src` folder?
For instance, the folders `.storybook`, `.next` or `docs` are not needed.

client/entrypoint.sh
davidsingal: To me is fine, but you can also add another line only for the build

client/Dockerfile
alexeh: @davidsingal thanks for your reply. As @tiagojsag stated, he has included tests stuff in the BE image, so IMO we either include tests stuff on the FE image, or remove it from BE container, up to you guys, but if they are **not needed**, I'd go for the latter.

Besides that @davidsingal, if you can kindly provide me with a list of the stuff you need in order to run the container, I'll make sure to include just that. No need to change your file tree bc this 😃

PR #29, opened 2021-06-22
chore: LANDGRIF-131 Create Suppliers module, Entity and entity_status enum

alexeh:
This PR Adds:

`Suppliers` nestjs module:
 -Entity model
 -Repository

`Entity Status` enumerable to be shared es enumerated type across different entities

`API_POSTGRES_PORT` en var in `compose` setup to expose the db service (for development purposes)

`API_DB_SYNCHRONIZE` en var to sync defined entities against the db on api init (on `dotenv default` and `compose` setup)


api/src/modules/suppliers/suppliers.entity.ts
```diff
@@ -0,0 +1,23 @@
...
+
+@Entity()
+export class Suppliers extends BaseEntity {
+  @PrimaryGeneratedColumn('uuid')
+  id: string;
+
+  @Column({ type: 'ltree', nullable: false, unique: true })
+  path: string;
+
+  @Column({ nullable: true })
```
tiagojsag, line 12: are we sure we want the name to be nullable?

alexeh: @tiagojsag  Right.  Fix -> fda24d1bbf475b29231af2b03860ac10fca012cb

Besides that, just followed the provided definition:
image
So I guess if its not marked as NOT NULL, then can be nullable

TypeORM,s default value for nullable is `false`, but just added it for better readability IMO

tiagojsag: we shouldn't take the model definition by Francis/Elene without adding our own input and experience to it. I'm sure they put a lot of work and care into it, but I'm sure they are open to our input on it. If we anticipate any problem with a particular definition, we can talk with them about it or, for minor things like this, we can just go with our suggestion, which can easily be modified in the future if needed anyways.

PR #33, opened 2021-06-23
LANDGRIF-136 - Controller and basic tests for BusinessUnits API

tiagojsag:


api/src/modules/business-units/business-units.controller.ts
```diff
@@ -0,0 +1,91 @@
...
+    @Param('id') id: string,
+  ): Promise<BusinessUnit> {
+    return await this.service.serialize(await this.service.update(id, dto));
+  }
+
+  @ApiOperation({ description: 'Deletes a business unit' })
+  @ApiOkResponse()
+  @Delete(':id')
+  async delete(
+    @Body(new ValidationPipe()) dto: UpdateBusinessUnitsDto,
```
alexeh, line 86: Is this dto validation required in deletion?

PR #38, opened 2021-06-25
Push docker image to dockerhub

tiagojsag:
- builds docker images for both FE and API for the main branch
- Published those images to dockerhub
- updates the cluster deployment with the new images
- configures the DNSs for the cluster
- Adds k8s deployments for the above apps
- Fixes a few things here and there
- With so much code, it probably mines bitcoin and you wont even notice it (I promise one day I'll make decent sized PRs)


.github/workflows/publish-docker-images.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Push to Docker Hub
+        uses: docker/build-push-action@v2
+        with:
+          context: api
+          push: true
+          tags: vizzuality/landgriffon-api:latest
+
+  push_client_to_registry:
+    name: Push Client Docker image to Docker Hub
+    runs-on: ubuntu-latest
```
alexeh, line 48: Evil:

```suggestion
    runs-on: ubuntu-20.04
```

.github/workflows/publish-docker-images.yml
```diff
@@ -0,0 +1,77 @@
...
+      - 'main'
+    paths:
+      - 'api/**'
+      - 'client/**'
+      - '.github/**'
+
+jobs:
+  push_api_to_registry:
+    name: Push API Docker image to Docker Hub
+    runs-on: ubuntu-latest
```
alexeh, line 15: Super-Evil
(probably btc mining part introduced here)

```suggestion
    runs-on: ubuntu-20.04
```

client/Dockerfile
```diff
@@ -19,14 +19,25 @@ COPY --chown=$USER:$USER package.json yarn.lock ./
...
-# Testing folders / files
-COPY --chown=$USER:$USER cypress cypress.json ./
-# Project config giles
-COPY --chown=$USER:$USER .* ./
-COPY --chown=$USER:$USER entrypoint.sh tsconfig.json ./
+COPY --chown=$USER:$USER next.config.js next-env.d.ts now.json \
+	postcss.config.js local.d.ts tailwind.config.js cypress.json entrypoint.sh \
+	tsconfig.json .babelrc .browserlistrc ./
+
+RUN yarn build
```
alexeh, line 41: Thanks for fixing this :)

docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
     ports:
       - "${CLIENT_SERVICE_PORT}:3000"
     container_name: landgriffon-client
-    command: develop
+    command: start
```
alexeh, line 11: This does not work (at least for local dev) because NextJS need a previous build (generates a `build` folder in the FS by `next build` and then runs it by `next start`.

`landgriffon-client | Error: Could not find a production build in the '/opt/landgriffon-client/.next' directory. Try building your app with 'next build' before starting the production server.`

This is the reason I intended to add a [single-command build&run](https://github.com/Vizzuality/landgriffon/pull/24#discussion_r649328722) but got rejected

alexeh (commented):
Thanks @tiagojsag for bringing CD to our lifes. I added some comments but overall I have no way to test in a functional way this branch's scope. Would be nice to have some kind of walkthrough, maybe in a devismos meeting

If the changes of the `compose` setup are required for this k8s deploy, I'd suggest to add a new `dev-compose` setup for local dev, and also nodemon for QoL improvements

docker-compose.yml
tiagojsag: I changed this back to develop.  I am also changing both entrypoint.sh files to more closely resemble each other, and to make them more explicit in their sub-commands, and thus require less brain power to manage. (this will go against the change I requested before, I know, sorry about it, but humans do change their minds sometimes).

The docker-compose is used exclusively for local dev, the only requirement is that it plays nice with the Dockerfile's we have for both apps - those are used to build the images that go to prod. If you want to modify this file to better suit your local dev env, be my guest.

alexeh (approved):

PR #39, opened 2021-06-28
generation of mock data

elpamart:
The PR contains a notebook with the generation of the different risk maps needed for deriving the mock data for the user locations and commodities, as well as the final metric (unsustainable water use, deforestation due to land use change, carbon emissions due to land use change and biodiversity loss due to land use change). In the same PR you will find a zip file with the final impact metrics and the different risk maps.

For getting the final results we have done a series of assumptions:

- we are using a temporal emissions datasets instead of the net flux from gfw
- we will need to replace the temporal deforestation datasets with the glad alerts from gfw and the sateligence data 
- for a first iteration we are avoiding the methodology of the probability area distribution. Therefore, all the result have been obtained for the different geometries and we will need to incorporate the probability area distribution methodology later on the process to get a heat map.


01painadam (approved):
Great job, the methodology here is well documented, easy to follow, and reproducible!

01painadam:
One small note, if we are doing a simple linear projection it might be useful to also include some uncertainty (using std dev) in the projection.

PR #44, opened 2021-07-02
feature: LANDGRIF-147 - Add endpoint for XLSX uploads w/ size limit + parsing

alexeh:
This PR adds:

-A endpoint in `Sourcing-Locations Controller` (TBC) to allow file uploading
-file-utils file with options for multer interceptor (storage, limits..)
-A custom decorator for Swagger docs
-A service to handle files inside the API
-A wrapper for parsing XLSX

It just allows to upload a file via POST request and stores it in `/tmp` folder. Plus size limit currently hardcoded to 1MB


tiagojsag (changes requested):
a few change requests:
- I think this would fit better in a module of its own. For now it's a very basic controller, but it will grow in the future.
- I don't see why we can't tackle those @debt straight away. I am ok with debt for complex things, but adding env vars doesn't sound like it
- a "finally {deleteFile()} would be great to ensure the FS stays clean
- Tests should NOT be a separate task

PR #45, opened 2021-07-02
Moved all the project to a src folder

davidsingal:
Next JS also supports the src folder. 
I moved everything project related to this folder because it was a bit hard to navigate and search.  I've also edited the Dockerfile.
Additionally, svgs has been renamed to assets.


alexeh:
Hi @davidsingal thanks for taking the time to re-order this stuff! 

I just checked that the whole thing gets built properly running via docker / docker-compose so LGTM

I also noticed that the client needs a bit more time to build pages (at least in dev mode) but this is probably due to my lack of computational resources, so if this works for you, works for me as well!

Thanks!

alexeh (approved):

pedrogpimenta (approved):
LGTM

PR #46, opened 2021-07-05
LANDGRIF-137 - Sourcing Record Groups CRUD

tiagojsag:


api/src/modules/sourcing-record-groups/dto/create.sourcing-record-group.dto.ts
```diff
@@ -0,0 +1,27 @@
...
+  @IsString()
+  @IsOptional()
+  @ApiPropertyOptional()
+  description?: string;
+
+  @IsString()
+  @IsOptional()
+  @IsJSON()
+  @ApiPropertyOptional()
+  metadata?: string;
```
alexeh, line 26: ?

```suggestion
  metadata?: JSON;
```

api/src/modules/sourcing-record-groups/sourcing-record-group.entity.ts
```diff
@@ -0,0 +1,58 @@
...
+    type: 'timestamp',
+    default: () => 'CURRENT_TIMESTAMP',
+  })
+  @ApiProperty()
+  lastEdited!: string;
+
+  @ManyToOne(() => User, (user: User) => user.sourcingRecordGroups, {
+    eager: false,
+  })
+  @ApiProperty()
```
alexeh, line 49: Each time you don't build the app after changes, your soul breaks and becomes an horcrux

```suggestion
 @ApiProperty({ type: () => User })
```
(Circular dependency issues make the api crash)

api/src/modules/sourcing-record-groups/sourcing-record-group.entity.ts
```diff
@@ -0,0 +1,58 @@
...
+  lastEditedUser?: User;
+
+  /**
+   * @debt Auto-assign user and make not nullable
+   */
+  @Column({ nullable: true })
+  @ApiPropertyOptional()
```
alexeh, line 56: Is this decorator necesary? Legit question

Which is the point to document `lastEditedUser` as  `@ApiProperty()` and then lastEditedUserId as `@ApiPropertyOptional()`?

tiagojsag: great legit question. once I have a legit answer, I'll tell you :P
I need to look into this at some point, once we double down on the API docs

alexeh: I already look it up for you 😝 
I'd say this makes no sense, because of that first `@ApiProperty()`
image

api/src/modules/sourcing-record-groups/sourcing-record-groups.entity.ts
```diff
@@ -0,0 +1,40 @@
+import {
```
alexeh, line 1: File duplicated? There are 2 `sourcing-record-groupsS` entities in this PR

tiagojsag: *facepalm*

PR #49, opened 2021-07-05
Landgrif-137   indicator sources crud

tiagojsag:

PR #50, opened 2021-07-05
Landgrif 137   scenarios crud

tiagojsag:


alexeh:
Hi! This PR includes changes from 

- https://github.com/Vizzuality/landgriffon/pull/46 

- https://github.com/Vizzuality/landgriffon/pull/49

So if you are ok with this, I'll review this one once these mentioned above are sorted out so I can avoid going mad 😄 

Thanks!

api/src/modules/scenarios/scenario.entity.ts
```diff
@@ -0,0 +1,89 @@
...
+   * @debt Auto-assign user and make not nullable
+   */
+  @Column({ nullable: true })
+  @ApiPropertyOptional()
+  userId?: string;
+
+  @ManyToOne(() => User, (user: User) => user.scenariosLastEdited, {
+    eager: false,
+  })
+  @ApiProperty()
```
alexeh, line 80: Again Circular Dep. isssue.

I'd fix this myself but not sure how you want to tackle it because we are following different appraches through entities

api/src/modules/scenarios/scenario.entity.ts
```diff
@@ -0,0 +1,89 @@
...
+  BaseEntity,
+  OneToMany,
+  ManyToOne,
+} from 'typeorm';
+
+import { SourcingLocation } from 'modules/sourcing-locations/sourcing-location.entity';
+
+import { BaseServiceResource } from 'types/resource.interface';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { User } from '../users/user.entity';
```
alexeh, line 14: ESLint does not complain?

api/src/modules/users/user.entity.ts
```diff
@@ -10,6 +10,7 @@ import { BaseServiceResource } from 'types/resource.interface';
 import { IndicatorCoefficient } from 'modules/indicator-coefficients/indicator-coefficient.entity';
 import { SourcingLocation } from 'modules/sourcing-locations/sourcing-location.entity';
 import { SourcingRecordGroup } from '../sourcing-record-groups/sourcing-record-group.entity';
```
alexeh, line 12: Relative imports are not allowed in our linting settings

PR #52, opened 2021-07-08
added collapsed button and show it in the URL [LANDGRIF-128]

davidsingal:


pedrogpimenta (changes requested):
Collapse button arrow should be the other way:
- When sidebar is open, chevron should point Left
- When sidebar is closed, chevron should point Right

pedrogpimenta (approved):

PR #54, opened 2021-07-12
LANDGRIDF-137 - Scenario intervention

tiagojsag:


alexeh:
@tiagojsag Thanks! I took the liberty to add some fixes, listing them here to keep track if needed:

-Fix circular dependency issue
-Add validation pipe to controller handlers
-Some eslint complains and minor semantic stuff
-Fix tests and make CI great again

Behaviour tested for this controller, ready to go

PS: You just created your fifth horcrux, please keep in mind that the max known amount of horcruxes is 6, after that no one knows what could happen....

PR #56, opened 2021-07-12
Add friendly error message on POST without required fields

tiagojsag:
I also did a full text reformat, so you may find unexpected changes, like import reorganization and such.


api/test/jest-custom-matchers.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    return {
+      pass: true,
+      message: (): string => `ok`,
+    };
+  },
+});
+
+export {};
```
alexeh, line 82: Really nice, thanks for this

alexeh (approved):

PR #57, opened 2021-07-12
Extract createdat and updatedat fields to separate base entity

tiagojsag:
Because a good engineer is a lazy engineer


api/src/baseEntities/timestamped-base-entity.ts
```diff
@@ -0,0 +1,16 @@
...
+  })
+  @ApiProperty()
+  createdAt!: string;
+
+  @UpdateDateColumn({
+    type: 'timestamptz',
+  })
+  @ApiProperty()
+  updatedAt!: string;
+}
```
alexeh, line 16: Also really nice, loved the idea

alexeh (approved):

PR #64, opened 2021-07-17
Add tree endpoint and parent children relationship management to material

tiagojsag:
addresses the last of https://vizzuality.atlassian.net/browse/LANDGRIF-189 I think

- Adds ability to pass parentId when creating a material
- optionally allows requesting children when loading materials list
- Adds new /material/tree endpoint to load data in a tree-like format
- Adds optional depth limiter on the tree endpoint (implementation will be reviewed but API contract wont change)


api/src/modules/materials/materials.controller.ts
```diff
@@ -62,6 +70,35 @@ export class MaterialsController {
     return this.materialsService.serialize(results.data, results.metadata);
   }
 
+  @ApiOperation({
+    description:
+      'Find all materials and return them in a tree format. Data in the "children" will recursively extend for the full depth of the tree',
+  })
+  @ApiOkResponse({
```
alexeh, line 77: Maybe a custom decorator would feel cleaner here

alexeh:
Thanks @tiagojsag, this is brilliant.

I have some questions that I'd like to clarify:

- Swagger shows that `depth` query is required for the request, which is not accurate. 

image

- Commented out tests: Why are these tests commented out? As an example of possible use cases regarding to this new functionality?

PR #74, opened 2021-07-23
Add analysis filters

pedrogpimenta:


client/src/containers/analysis-visualization/analysis-filters/group-by/index.tsx
```diff
@@ -0,0 +1,25 @@
+import type { Group } from 'types';
+import Component from './component';
+
+const ALL_GROUPS: Group[] = [
+  {
+    id: '0',
```
davidsingal, line 6: @pedrogpimenta I'd prefer to use a different index. Something like 'material', because it's what we are going to use in the URL and it will be more friendly.

PR #75, opened 2021-07-24
BUGFIX: Sourcing Data fixes

alexeh:
**PLEASE, READ** (It's actually sad I have to put this here)

This PR Fixes:

- Incorrect meta / non-meta prop buildin. Cannot explain when this happened, must have been a dark elf

- Lost coordinates: Lodash's `isEmpty` always evaluating the condition to true, therefore `lat` and `long` always `undefined`

- Little changes in the `sourcing-data building algorithm - dto parsing` to have awareness of `sourcing-location -> sourcing-records relation
`

**KEY NOTES**

-As I already mentioned, I'd say we should try to change N-1 relation to 1-N  to decrease time complexity, and also increase our QoL letting the ORM do the dirty job. I believe I have no other way to know the relation between involved entities in the way that the data comes to us

-Also I think that 'forcing' to validate data against Api-Client contract DTOs, when data-shape before persisting point does not match, does not make any sense.  The `entity definition` itself is the contract between API and persistence layer, so we could either:
  - Use the `entity` itself for both validation and typing, but this would imply to add validation decorators to the entity, and could get messy to read
  - Create specific DTOs for API-DB trades, but this also sounds a bit nasty. I'd rather go for the former

### DIFF
--- a/api/src/modules/import-data/sourcing-records/dto-processor.service.ts
+++ b/api/src/modules/import-data/sourcing-records/dto-processor.service.ts
@@ -8,9 +8,10 @@ import { CreateSourcingRecordDto } from 'modules/sourcing-records/dto/create.sou
 import { CreateSourcingLocationDto } from 'modules/sourcing-locations/dto/create.sourcing-location.dto';
 import { Layer } from 'modules/layers/layer.entity';
 import { WorkSheet } from 'xlsx';
+import { SourcingRecord } from 'modules/sourcing-records/sourcing-record.entity';
 
 export interface SourcingData extends CreateSourcingLocationDto {
-  sourcingRecords: CreateSourcingRecordDto[];
+  sourcingRecords: SourcingRecord[];
 }
 
 export interface SourcingRecordsDtos {
--- a/api/src/modules/import-data/sourcing-records/dto-processor.service.ts
+++ b/api/src/modules/import-data/sourcing-records/dto-processor.service.ts
@@ -10,6 +10,10 @@ import { Layer } from 'modules/layers/layer.entity';
 import { WorkSheet } from 'xlsx';
 import { SourcingRecord } from 'modules/sourcing-records/sourcing-record.entity';
 
+/**
+ * @debt: Define a more accurate DTO / Interface / Class for API-DB trades
+ * and spread through typing
+ */
 export interface SourcingData extends CreateSourcingLocationDto {
   sourcingRecords: SourcingRecord[];
 }
--- a/api/src/modules/import-data/sourcing-records/dto-processor.service.ts
+++ b/api/src/modules/import-data/sourcing-records/dto-processor.service.ts
@@ -6,17 +6,19 @@ import { CreateAdminRegionDto } from 'modules/admin-regions/dto/create.admin-reg
 import { SourcingRecordsSheets } from 'modules/import-data/sourcing-records/import.service';
 import { CreateSourcingRecordDto } from 'modules/sourcing-records/dto/create.sourcing-record.dto';
 import { CreateSourcingLocationDto } from 'modules/sourcing-locations/dto/create.sourcing-location.dto';
-import { isEmpty } from 'lodash';
 import { Layer } from 'modules/layers/layer.entity';
 import { WorkSheet } from 'xlsx';
+
+export interface SourcingData extends CreateSourcingLocationDto {
+  sourcingRecords: CreateSourcingRecordDto[];
+}
 
 export interface SourcingRecordsDtos {
   materials: CreateMaterialDto[];
   adminRegions: CreateAdminRegionDto[];
   businessUnits: CreateBusinessUnitDto[];
   suppliers: CreateSupplierDto[];
-  sourcingRecords: CreateSourcingRecordDto[];
-  sourcingLocations: CreateSourcingLocationDto[];
+  sourcingData: SourcingData[];
 }
 
 const NON_META_PROPERTIES: Array<string> = [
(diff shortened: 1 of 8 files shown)


tiagojsag (changes requested):
Please add some tests to illustrate the problems you're trying to fix here.

api/test/import-data/xlsx-uploads/sourcing-records-import.spec.ts
```diff
@@ -170,4 +170,20 @@ describe('Sourcing Records import', () => {
     const sourcingRecords: SourcingRecord[] = await sourcingRecordRepository.find();
     expect(sourcingRecords.length).toEqual(825);
   });
+
```
alexeh, line 173: Test for the fix

api/src/modules/sourcing-records/sourcing-record.entity.ts
```diff
@@ -35,10 +35,14 @@ export class SourcingRecord extends TimestampedBaseEntity {
   @ApiProperty()
   year!: number;
 
-  @ManyToOne(() => SourcingLocation, (srcLoc: SourcingLocation) => srcLoc.id)
-  @JoinColumn()
-  @ApiPropertyOptional()
-  sourcingLocationsId: string;
+  @ManyToOne(() => SourcingLocation, (srcLoc: SourcingLocation) => srcLoc.id, {
```
tiagojsag, line 38: no need to save on keystrokes, keyboards are cheap. Use descriptive var names.

api/src/modules/sourcing-records/sourcing-record.entity.ts
```diff
@@ -49,5 +53,5 @@ export class SourcingRecord extends TimestampedBaseEntity {
    */
   @ManyToOne(() => User, (user: User) => user.id)
   @ApiProperty()
-  updatedById?: string;
+  updatedBy?: string;
```
tiagojsag, line 56: yeah, we have a few of these we need to modify. good catch

api/src/modules/sourcing-records/sourcing-record.entity.ts
```diff
@@ -35,10 +35,14 @@ export class SourcingRecord extends TimestampedBaseEntity {
...
-  @ApiPropertyOptional()
-  sourcingLocationsId: string;
+  @ManyToOne(() => SourcingLocation, (srcLoc: SourcingLocation) => srcLoc.id, {
+    onDelete: 'CASCADE',
+  })
+  @JoinColumn({ name: 'sourcingLocationId' })
+  sourcingLocation: SourcingLocation;
+
+  @Column({ nullable: true })
+  sourcingLocationId: string;
```
tiagojsag, line 45: add  the !

api/test/import-data/xlsx-uploads/sourcing-records-import.spec.ts
```diff
@@ -27,9 +27,10 @@ describe('Sourcing Records import', () => {
   let adminRegionRepository: AdminRegionRepository;
   let sourcingLocationRepository: SourcingLocationRepository;
   let sourcingRecordRepository: SourcingRecordRepository;
-  let sourcingRecordGroupRepository: SourcingLocationGroupRepository;
+  let sourcingLocationGroupRepository: SourcingLocationGroupRepository;
 
   beforeAll(async () => {
+    jest.setTimeout(10000);
```
tiagojsag, line 33: no, this is evil AF. set it in jest-config.json

api/test/import-data/xlsx-uploads/sourcing-records-import.spec.ts
```diff
@@ -170,4 +170,20 @@ describe('Sourcing Records import', () => {
...
+      .attach('file', __dirname + '/base-dataset.xlsx');
+
+    const sourcingRecords: SourcingRecord[] = await sourcingRecordRepository.find();
+    const sourcingLocation:
+      | SourcingLocation
+      | undefined = await sourcingLocationRepository.findOne(
+      sourcingRecords[0].sourcingLocationId,
+    );
+
+    expect(sourcingRecords[0].sourcingLocationId).toBeTruthy();
```
tiagojsag, line 186: do proper type validation here

api/test/import-data/xlsx-uploads/sourcing-records-import.spec.ts
```diff
@@ -170,4 +170,20 @@ describe('Sourcing Records import', () => {
...
+    expect(sourcingLocation).toBeDefined();
```
tiagojsag, line 187: also do proper validation here, to ensure you are actually getting something that is a sourcingLocation.

tiagojsag (changes requested):

api/src/modules/import-data/sourcing-records/dto-processor.service.ts
```diff
@@ -6,17 +6,19 @@ import { CreateAdminRegionDto } from 'modules/admin-regions/dto/create.admin-reg
 import { SourcingRecordsSheets } from 'modules/import-data/sourcing-records/import.service';
 import { CreateSourcingRecordDto } from 'modules/sourcing-records/dto/create.sourcing-record.dto';
 import { CreateSourcingLocationDto } from 'modules/sourcing-locations/dto/create.sourcing-location.dto';
-import { isEmpty } from 'lodash';
 import { Layer } from 'modules/layers/layer.entity';
 import { WorkSheet } from 'xlsx';
 
+export interface SourcingData extends CreateSourcingLocationDto {
```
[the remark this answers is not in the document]
alexeh: @tiagojsag, yeah this is the reason of my comment in the PR description. Let's clarify it and implement the 'best' solution

api/src/modules/import-data/sourcing-records/import.service.ts
```diff
@@ -85,11 +85,13 @@ export class SourcingRecordsImportService {
   ): Promise<void | Array<ErrorConstructor>> {
     const validationErrorArray: Array<typeof Error> = [];
     for (const parsedSheet in dtoLists) {
-      for (const dto of dtoLists[parsedSheet as keyof SourcingRecordsDtos]) {
-        try {
-          await validateOrReject(dto);
-        } catch (err) {
-          validationErrorArray.push(err);
+      if (dtoLists.hasOwnProperty(parsedSheet)) {
```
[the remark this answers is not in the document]
alexeh: @tiagojsag  the purpose of this clause is to avoid the `for of` loop to look down through all the prototype chain, which we totally sure we don't want / need

api/src/modules/import-data/sourcing-records/import.service.ts
```diff
@@ -120,4 +122,31 @@ export class SourcingRecordsImportService {
...
+   *
+   * @param sourcingData: Array of Sourcing-Locations with nested Sourcing-Records
+   * @private
+   * Saves each Sourcing Location and returns and ID to attach to its related
+   * Sourcing-Records, to keep relation
+   * Save Sourcing-Record array (ORM will manage better than us) with attached Sourcing-Location ID
+   * @debt: There MUST be a cleaner and more performant way to do this
+   */
+
+  private async saveSourcingData(sourcingData: SourcingData[]): Promise<void> {
```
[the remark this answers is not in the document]
alexeh: @tiagojsag totally true, brilliant catch.

api/test/import-data/xlsx-uploads/sourcing-records-import.spec.ts
alexeh: @tiagojsag IMO we need this timeOut increase just in these XLSX uploads tests so far, as these are 'heavy' ops, no need for this timeOut increase to be applied to the rest of the tests (Mostly simple CRUDs)

Applying them for all test cases could hide some sign of sickness in the flow

tiagojsag: fair point, but then you need to reset the previous value on after()

PR #76, opened 2021-07-28
Feature: LANDGRIF-146 - Allow CRUD to link entities

alexeh:
This PR adds:

Capabilities for retrieving relational entities 
**Currently for the data for we know / have relationships**

ON DELETE CASCADE for N->1 relations, again, for the entities we have relations for
**Because allowing inconsistent data will lead us to pain and suffering**

Some tests

**NOTE**
There are still missing pieces for data and relations, I worked with the info I have right now
We are allowing nullable FK when we shouldn't, but it is what it is

**There will be changes here, but FE can start retrieving related stuff**

### DIFF
--- a/api/src/modules/sourcing-locations/sourcing-location.entity.ts
+++ b/api/src/modules/sourcing-locations/sourcing-location.entity.ts
@@ -37,7 +37,7 @@ export const sourcingLocationResource: BaseServiceResource = {
     singular: 'sourcingLocation',
     plural: 'sourcingLocations',
   },
-  entitiesAllowedAsIncludes: [],
+  entitiesAllowedAsIncludes: ['sourcingLocationGroup'],
   columnsAllowedAsFilter: ['title'],
 };
 
@@ -101,16 +101,31 @@ export class SourcingLocation extends TimestampedBaseEntity {
   @ManyToOne(() => User, (user: User) => user.sourcingLocations, {
     eager: false,
   })
+  @JoinColumn({ name: 'updatedById' })
   updatedBy: User;
+
+  /**
+   * @debt: Make this required and auto-set once auth is great again
+   */
+  @Column({ nullable: true })
+  updatedById: string;
 
   @ManyToOne(() => Material, (mat: Material) => mat.sourcingLocations, {
     eager: false,
+    onDelete: 'CASCADE',
   })
+  @JoinColumn({ name: 'materialId' })
   material: Material;
+
+  @ApiPropertyOptional()
+  @Column({ nullable: true })
+  materialId: string;
 
   @ManyToOne(() => AdminRegion, (aR: AdminRegion) => aR.sourcingLocations, {
     eager: false,
+    onDelete: 'CASCADE',
   })
+  @JoinColumn({ name: 'adminRegionId' })
   adminRegion: AdminRegion;
 
   @Column({ nullable: true })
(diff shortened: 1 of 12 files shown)


api/src/modules/sourcing-locations/sourcing-location.entity.ts
```diff
@@ -119,14 +134,21 @@ export class SourcingLocation extends TimestampedBaseEntity {
 
   @ManyToOne(() => BusinessUnit, (bu: BusinessUnit) => bu.sourcingLocations, {
     eager: false,
+    onDelete: 'CASCADE',
```
tiagojsag, line 137: I wouldnt cascade delete here.

api/src/modules/sourcing-locations/sourcing-location.entity.ts
```diff
@@ -119,14 +134,21 @@ export class SourcingLocation extends TimestampedBaseEntity {
...
+  @ApiPropertyOptional()
+  @Column({ nullable: true })
+  businessUnitId: string;
+
   @ManyToOne(
     () => Supplier,
     (supplier: Supplier) => supplier.sourcingLocations,
     {
       eager: false,
+      onDelete: 'CASCADE',
```
tiagojsag, line 151: same here, I wouldnt cascade delete

api/src/modules/sourcing-locations/sourcing-location.entity.ts
```diff
@@ -136,6 +158,7 @@ export class SourcingLocation extends TimestampedBaseEntity {
...
```
tiagojsag, line 161: and here

tiagojsag (changes requested):
Minor stuff, otherwise LGTM

api/src/modules/sourcing-locations/sourcing-location.entity.ts
alexeh: @tiagojsag this is not my whim, after having a chat with Elena it was clear that a SourcingLocations needs a Business-Unit, always. Is there any reason to keep this? We could set a default value for businessUnitId in case that Business Unit gets deleted, but I think we'd be going against business-logic. This also could affect the 10000000 filters FE is implementing

api/src/modules/sourcing-locations/sourcing-location.entity.ts
alexeh: [Same reason here](https://github.com/Vizzuality/landgriffon/pull/76#discussion_r679136595)

api/src/modules/sourcing-locations/sourcing-location.entity.ts
alexeh: [Same reason here](https://github.com/Vizzuality/landgriffon/pull/76#discussion_r679136595)

api/src/modules/sourcing-locations/sourcing-location.entity.ts
tiagojsag: so I understood that what cascade meant here was that if you deleted the sourcing location, the business unit would be deleted too. Stackoverflow says otherwise (which makes more sense anyways) so if that's the case, you are right, and it should be left as it is

alexeh: @tiagojsag the other way around. Same for Sourcing-Records and Sourcing-Locations. If a Sourcing-Location gets deleted, all its 'dependant' Sourcing-Records will be deleted to

PR #77, opened 2021-07-28
Raster to h3 table import script

fgassert:
This PR adds a dockerized script to download a raster dataset, convert it to h3, and import it into the database.

The script is documented at `data/h3_raster_importer/README.md`

### Jira
This is a proof of concept for: https://vizzuality.atlassian.net/browse/LANDGRIF-207

### Test
Follow the instructions to run the script at `data/h3_raster_importer/README.md`
You should end up with a table `h3_grid_spam2010v2r0_global_prod` with 43 columns, one for each crop in the database.


data/h3_raster_importer/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    landgriffon-seed-data:
+        build:
+            context: ./
+            dockerfile: Dockerfile
+        volumes:
+            - '../data/seed:/data'
+        command: make
+        env_file:
+            - '../../.env'
+        environment:
```
alexeh, line 12: Are these env vars required? 
I think you are getting values straight from the env file via

`psycopg2.connect(
        host=os.getenv('API_POSTGRES_HOST'),
        port=os.getenv('API_POSTGRES_PORT'),
        user=os.getenv('API_POSTGRES_USERNAME'),
        password=os.getenv('API_POSTGRES_PASSWORD')
    )`
    
    I don't see them anywhere else, but I could be missing something

data/h3_raster_importer/tiff_folder_to_h3_table.py
```diff
@@ -0,0 +1,145 @@
...
+
+def load_raster_list_to_h3_table(raster_list, table, h3_res):
+    conn = psycopg2.connect(
+        host=os.getenv('API_POSTGRES_HOST'),
+        port=os.getenv('API_POSTGRES_PORT'),
+        user=os.getenv('API_POSTGRES_USERNAME'),
+        password=os.getenv('API_POSTGRES_PASSWORD')
+    )
+    cursor = conn.cursor()
+    cursor.execute(f"DROP TABLE IF EXISTS {table};")
```
alexeh, line 105: I believe the API should take care of the existence of the table, as it already takes care of generating the full data-model on the DB at init, also this is relevant for us in order to access the data stored on said table

data/h3_raster_importer/tiff_folder_to_h3_table.py
```diff
@@ -0,0 +1,145 @@
...
+    
+    first = True
+
+    for block_df in gen_raster_h3(raster_list, h3_res):
+        if first:
+            cols = zip(block_df.columns, block_df.dtypes)
+            schema = ', '.join([f"{col} {DTYPES_TO_PG[str(dtype)]}" for col, dtype in cols])            
+            cursor.execute(f"CREATE TABLE {table} (h3index h3index PRIMARY KEY, {schema});")
+            conn.commit()
+            logging.info(f"Created table {table} with columns {', '.join(block_df.columns)}")
```
alexeh, line 115: I'd say you are generating the table and the columns depending on downloaded data, but both table and column names are relevant from the API point of view in order to access this data. Could we define some naming or make it match [the data model](https://dbdiagram.io/d/60bf8dd0b29a09603d1883ef)?

alexeh:
Thank you @fgassert for taking the time for this. I just tested it locally and table is created and data is imported as expected.
Nevertheless, I have some questions / doubts about it. I left some comments that I'd need to clarify

I am not sure if this is a temporary solution / patch for us to keep going forward, or it is a permanent solution, but I'd love to have a chat about it, since this kind of stuff (h3 etc) is totally new for me and I'm probably missing parts here

Thanks!

data/h3_raster_importer/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM perrygeo/gdal-base
+
+RUN pip install --upgrade --no-cache-dir pip
+RUN pip install --no-cache-dir numpy pandas rasterio h3ronpy psycopg2-binary docopt
```
tiagojsag, line 4: every time you install python dependencies without pinning the versions, you-in-6-months kills a kitty because a new version (with breaking changes) was introduced, and your code magically stopped working.

data/h3_raster_importer/docker-compose.yml
tiagojsag: I don't recall how this works, but I suspect you may be right. we should test this, and remove the "environment" section if it's not needed

data/h3_raster_importer/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: "3.8"
+services:
+    landgriffon-seed-data:
+        build:
```
tiagojsag, line 4: pro tip: these 3 lines can be simplified to "build: ."

data/h3_raster_importer/Dockerfile
tiagojsag, line 4: every time you install python dependencies without pinning the versions, you-in-6-months kills a kitty because a new version (with breaking changes) was introduced, and your code magically stopped working.

tiagojsag (changes requested):
I second Alex's comment on this script manipulating the database structure. Dropping and creating tables in a vacuum is potentially evil (what happens to constraint/FKs/other related stuff?) so we should avoid doing it in this script. I would suggest this script focuses on validating that the table exists and that it has the agree upon structure. If it does, insert the data (maybe even truncate the table before). If it doesn't, fail fast and send an email to the backend team.

fgassert:
Haha good. I was hoping to have this conversation. I agree with you in principle, and in the short term it would even make things easier. But over the long term, I think we want the GIS table(s) and their schemas to be flexible.

I'm think we should follow a pattern similar to how PostGIS handles rasters or even how CARTO handles user data. Essentially, these tables are more akin to binary blobs than entities (its only by making them into tables, we get to take advantage of postgres magic). There is then another nicely structured table that tracks all of these GIS data tables, which we can use to track FK relations etc.

tiagojsag:
The goal here is not to hinder flexibility, it's about centralizing as much as we can under a single source of truth. 

My opinion is that, if we're going to build an API endpoint that that serves this data, it will be closely tied to the data structure we have. Also, we already have TypeORM as the DB/data format management for the rest of the tables, so this is why I think we should use it also to maintain the structure. Note that this does not mean we need an entity-like table - I'm happy with an "id + blob" table if that's what makes sense.

Ultimately what I see as the best solution is having as much "truth" as possible on TypeORM's side, and this script being as barebones as it can be - so for example, it can make a lot of assumptions (db exists, table exists, has a given structure) and if any of them fails, the script itself fails fast, instead of trying to correct them. This makes this script easier to maintain, while also ensuring we avoid any sort of problems from having potentially 2 sources of truth.

As for flexibility, moving this truth ownership into the API code does add a barrier to you, in case you want to modify the table structure. So if you need to change something, we commit to helping you with it on the app side, so we can continue to be flexible as the project evolves.

tiagojsag:
just made a small change to decouple this app from docker. By having python deps defined inside the dockerfile, we are making native setup a bit harder, so we should have the deps defined separately from the dockerfile

fgassert:
@tiagojsag I've addressed your changes, thanks!

@alexeh I've updated the script to add rows to the H3 master table. I also updated the table to have one row per h3_grid column, so you should get 42 rows each corresponding to an individual crop or crop group.

PR #78, opened 2021-07-30
add script gadm 3.6 import

elpamart:
add script for importing a simplified version of gadm geometries into the database (level_0, level_1 and level_2). This script has been added in a `data_download ` folder which contains different subfolders for the different data that we will need to import to the database and the corresponding makefiles (e.g. gadm_3.6 in this case). We can have a generic makefile in the data_download folder that imports all the data at once (happy to discuss this)

#### Test:

for testing go to the data folder and run "make seed-data". As result you should get three new tables in the database with the gadm geometries for each level.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -36,7 +36,6 @@
     "config": "^3.3.6",
     "helmet": "^4.6.0",
     "jsonapi-serializer": "^3.6.7",
-    "mapshaper": "^0.5.62",
     "ms": "^2.1.3",
     "multer": "^1.4.2",
     "nestjs-base-service": "0.7.1",
--- a/api/package.json
+++ b/api/package.json
@@ -36,6 +36,7 @@
     "config": "^3.3.6",
     "helmet": "^4.6.0",
     "jsonapi-serializer": "^3.6.7",
+    "mapshaper": "^0.5.62",
     "ms": "^2.1.3",
     "multer": "^1.4.2",
     "nestjs-base-service": "0.7.1",
--- a/data/data_download/Dockerfile
+++ b/data/data_download/Dockerfile
@@ -1,0 +1,15 @@
+FROM osgeo/gdal:ubuntu-full-3.2.1
+LABEL maintainer="hello@vizzuality.com"
+ENV NODE_OPTIONS=--max_old_space_size=16384
+RUN apt-get update \
+    && curl -sL https://deb.nodesource.com/setup_13.x | sh - \
+    && apt-get install -y --no-install-recommends nodejs make jq \
+    && apt-get install -y --no-install-recommends \
+           postgresql postgresql-client \
+           iputils-ping \
+           time \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN npm i -g mapshaper@0.5.34
+
+COPY ./ ./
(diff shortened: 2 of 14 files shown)


fgassert:
@alexeh this is ready for review.
I've rebased to main and added an SQL script to populate the `admin_region` and `geo_region` tables from the gadm data. It's a little messy, so let me know if you can think of a better way to do this.

api/package.json
```diff
@@ -36,6 +36,7 @@
     "config": "^3.3.6",
     "helmet": "^4.6.0",
     "jsonapi-serializer": "^3.6.7",
+    "mapshaper": "^0.5.62",
```
alexeh, line 39: I'd say this is not required in the API side, since we are not (yet) using it, but its being downloaded and used in your magical script 😃

data/data_download/gadm_3.6/populate_admin_regions.sql
```diff
@@ -0,0 +1,42 @@
...
+-- so we use a WHERE clause instead of an ON CONFLICT
+WHERE NOT EXISTS (SELECT 1 FROM admin_region a WHERE a.mpath = gadm_levels0_2.mpath);
+-- ON CONFLICT (mpath) DO UPDATE SET
+--     "geoRegionId" = EXCLUDED."geoRegionId";
+
+-- 3. Add parentIds since these are tracked separately from the path
+UPDATE admin_region child
+SET "parentId" = parent.id
+FROM admin_region parent
+WHERE subpath(child.mpath::ltree, 0, -1)::text = parent.mpath;
```
alexeh, line 40: I had to add:
```
CREATE EXTENSION IF NOT EXISTS ltree;
```
In order to make this work, I want to belive you had to install it as well in your local instance to make it work, otherwise I am missing something, but if its required, we should add it in this script as well
(and drop it at the end if its just required by this script)

alexeh:
Thank you so much @fgassert for your effort here. After making both of my laptops melt, I've been able to complete the process and seeing admin_regions and geo_regions populated has been such a relief 😄 

Although there are little missing pieces which I pointed you via slack, let's have a quick chat whenever you are available and see if we can properly close this

Thank you!

api/package.json
tiagojsag: and I strongly suspect that the changes to the yarn.lock are not just this package, so review those too please

data/data_download/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM osgeo/gdal:ubuntu-full-3.2.1
+LABEL maintainer="hello@vizzuality.com"
+ENV NODE_OPTIONS=--max_old_space_size=16384
+RUN apt-get update \
+    && curl -sL https://deb.nodesource.com/setup_13.x | sh - \
```
tiagojsag, line 5: Please split these lines into multiple RUN commands.

Run all apt-get install in a single command

Also: any reason why you are using a dead version of nodejs? if not, please use the same major version as the API/FE apps

data/data_download/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+    && apt-get install -y --no-install-recommends nodejs make jq \
+    && apt-get install -y --no-install-recommends \
+           postgresql postgresql-client \
+           iputils-ping \
+           time \
+    && rm -rf /var/lib/apt/lists/*
```
tiagojsag, line 11: why is this needed?

tiagojsag (changes requested):

fgassert:
@tiagojsag I've addressed your comments and rebased to main.

@alexeh I've added the ISOa2 values, updated the mpath to use the ids, and added a field for the gadmId.

It still takes a long time to run, the longest step is generating the h3indexes from the geometries, but that should only have to happen once.

PR #85, opened 2021-09-01
front(feat): analysis table

anamontiaga:
### Overview

This PR includes a basic Antd table component with custom less styles.
[Antd table documentation](https://ant.design/components/table)

### Designs

[Figma](https://www.figma.com/file/tDHyfnWfMsydWmKcXaX8dV/LandGriffon-tool?node-id=1035%3A4406)


client/src/containers/analysis-visualization/component.tsx
```diff
@@ -25,61 +73,8 @@ const AnalysisVisualization = () => {
...
-                          {impactFactor['2023']}
-                        </td>
-                      </tr>
-                    ))}
-                  </tbody>
-                </table>
-              </div>
-            </div>
-          </div>
+        <div className="flex p-6 pl-12 mt-16 left-12" style={{ maxWidth: '1800px' }}>
```
davidsingal, line 76: I think we can safely remove this `maxWidth`, in any case, it should be a percentage. Because it should grow when the user collapses the sidebar.

client/src/containers/analysis-visualization/table/mock.tsx
```diff
@@ -0,0 +1,332 @@
+const DATA = [
```
davidsingal, line 1: You couldn't know how the data will be, It will be more like:

```
const DATA = [
  {
    id: 'a1',
    indicator: 'Carbon emissions',
    values: [
      { year: 2021, value: 200, calculated: false },
      { year: 2022, value: 300, calculated: true } // data projected for non-existing years
    ],
    children: [
      { name: 'Material name', values: [{ year: 2021, value: 100  }]  }
    ]
  }
];
```

For the children, given the indicator id, we will request the values inside in a separate request. For now, I'm happy to include in an object `children` as you are already doing it.

davidsingal (changes requested):

client/src/containers/analysis-visualization/component.tsx
```diff
@@ -1,18 +1,66 @@
 import { useAppSelector } from 'store/hooks';
 import { analysis } from 'store/features/analysis';
 import Map from 'components/map';
+import Table from './table';
+import Title from './table/title';
```
mbarrenechea, line 5: Name it TableTitle please

mbarrenechea (changes requested):
I think we should extract all the logic related to the table to another container to avoid having a huge "analysis-visualization" one. Let's create analysis-table, analysis-map and analysis-chart so this component will act as a parent of the others and it will be only in charge of changing the visualization

davidsingal:
As we are adding a lot of changes every day I strongly recommend rebasing from the main branch at the beginning of the day.
Also, just let you know I'm going to merge this version. It's good enough and we can add what is left later.

PR #87, opened 2021-09-03
Proof of Concept: LANDGRIF-253 Retrieve H3 Indexes by Resolution

alexeh:
**NOTE: This PoC has been functionally approved**

This PR adds:

-Implementation to retrieve dynamically generated h3Indexes "sorted" by resolution, to address potential performance issues
 due to huge amount of data
 
 -Currently we are storing H3 Indexes at max resolution (6) so depending on the zoom level, FE does not need the full data (+1.6M rows)
 
 Hopefully the implemented test set will give you enough understanding of why / what has been done.
 
Enjoy!


api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -30,4 +30,49 @@ export class H3DataRepository extends Repository<H3Data> {
...
+      .where('h3.id = :id', { id: h3Id })
+      .getRawOne();
+
+    if (!h3Info)
+      throw new NotFoundException(
+        `Requested H3 with ID: ${h3Id} could not been found`,
+      );
+    try {
+      const h3data = await this
+        .query(` select h3parent.h3_to_parent, sum(${h3Info.h3columnName})  from
```
tiagojsag, line 62: Couldn't this be simplified as :

`
            select h3_to_parent(h3index, ${resolution}) as h3_to_parent, sum(${h3Info.h3columnName})
            from ${h3Info.h3tableName} hgsvrgp
            group by h3_to_parent
`

api/test/h3-data/mocks/create-fake-h3-data.ts
```diff
@@ -1,16 +1,28 @@
...
   h3ColumnName: string,
-): Promise<void> => {
+  bulk?: boolean,
+): Promise<string> => {
   await getManager().query(
     `CREATE TABLE ${h3TableName} (h3index h3index, ${h3ColumnName} float4);` +
       `INSERT INTO ${h3TableName} (h3index, ${h3ColumnName} ) VALUES ('861203a4fffffff', 1000);` +
       `INSERT INTO h3_data ("h3tableName", "h3columnName", "h3resolution") VALUES ('${h3TableName}', '${h3ColumnName}', 6 );`,
   );
+  if (bulk) {
```
tiagojsag, line 14: Fun exercise 1: implement this using Promise.all()
Fun exercise 2: implement this using a single query
Fun exercise 3: explain why this particular implementation sucks

api/test/h3-data/mocks/h3-fixtures.ts
```diff
@@ -0,0 +1,401 @@
+export const h3Fixtures = {
```
tiagojsag, line 1: I suspect that, for the purpose of testing, we should probably have some values different than 0, otherwise the result of queries will always be 0, no?

api/test/h3-data/h3-data.spec.ts
```diff
@@ -56,9 +57,49 @@ describe('H3-Data Module (e2e)', () => {
...
+      );
+      expect(Object.keys(response.body).length).toEqual(400);
+    });
+
+    test('When I query H3 data at minimal resolution, then I should get 8 h3indexes', async () => {
+      const id = await createFakeH3Data(fakeTable, fakeColumn, true);
+      const response = await request(app.getHttpServer()).get(
+        `/api/v1/h3/data/${id}?resolution=1`,
+      );
+      expect(Object.keys(response.body).length).toEqual(8);
```
tiagojsag, line 102: Given that you know the input data, you can probably validate the actual values here.

api/test/h3-data/h3-data.spec.ts
```diff
@@ -56,9 +57,49 @@ describe('H3-Data Module (e2e)', () => {
...
+      expect(response.body.errors[0].title).toEqual(
+        `Requested H3 with ID: ${FAKE_UUID} could not been found`,
+      );
+    });
+    test('When I query H3 data with no resolution provided, then I should get I should get each available h3indexes (max resolution)', async () => {
+      const id = await createFakeH3Data(fakeTable, fakeColumn, true);
+      const response = await request(app.getHttpServer()).get(
+        `/api/v1/h3/data/${id}`,
+      );
+      expect(Object.keys(response.body).length).toEqual(400);
```
tiagojsag, line 94: See my comment below about "coloring" the sample data. Once you do that, you can do some logic validation of those 400 values. For example, you can count the number of values that are 0 and assert that. Go from there and add some logic validation, to ensure you are not returning 'potato' 400 times

api/src/modules/h3-data/dto/h3-by-resolution.dto.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * DTO for querying a resolution given a H3 ID
+ */
+import { Type } from 'class-transformer';
+import { IsEnum, IsNumber, IsOptional } from 'class-validator';
+
+/**
+ * Default (and max) resolution to retrieve h3 indexes if none is given
+ */
+export const DEFAULT_RESOLUTION = 6;
```
tiagojsag, line 10: could you explain why we went for the value that (AFAIK) produces the most data? Wouldn't it be more reasonable to a) use the cheapest default, or b) make the res mandatory on the request? (I actually prefer b over a TBH)

tiagojsag (commented):
Overall looks good, just a few notes

api/test/h3-data/mocks/h3-fixtures.ts
alexeh: Actually 0 is a valid value for H3 Data. What we care the most are the keys of the resultant object, which are the actual H3 Indexes. Would't make a difference setting some random values

api/test/h3-data/h3-data.spec.ts
alexeh: The point of just counting the rows its because the magic here is not about the **values**, since these are persisted by an external dataset. 
The magic here is calculating the parent h3indexes, which are the keys in the retrieved data, that's why I didn't care about the values here. And since this use cases brings max resolution, this should return the full fixture

Thoughts?
PS: Anyways I see value in your suggestion of matching the results for the above case of calculating by resolution

api/test/h3-data/mocks/create-fake-h3-data.ts
alexeh: Answer for Fun exercise 3: It suck in terms of performance because it makes as much calls as keys in the fixture object are. It was the fastest and easiest way to do it, and since its a test, I thought this could be valid.
Best solution is to make it work in a single query / transaction since its perform a single call and DB takes care of performance in its own realm

Promise.all implementation is not actually THAT essential here since this uses a for of loop, and it actually awaits for async calls, which is not happening in map or forEach loops 

I'll implement this in a single query, cheers

api/src/modules/h3-data/dto/h3-by-resolution.dto.ts
alexeh: Because max resolution implies the full map, retrieving the cheapest result will imply returning the max zoom, which I am not sure if it makes sense. 
I totally see your point and I do agree with it, I thought, without asking anyone and thats not ok, making default the full data was the right answer.

Let me check with FE which would the solution will them

api/src/modules/h3-data/h3-data.repository.ts
alexeh: Yes, brilliant, we love KISS.
Thanks!

alexeh:
Thank you @tiagojsag for reviewing this PR. Suggestions applied:

1. Fixed config param for GMAPS_API_KEY
2. Simplified SQL query for calculate parent H3 Indexes
3. Make resolution param required in the request - test updated accordingly
4. Test stuff improved (And Fun exercises completed hopefully with the highest grade :) )

PR #88, opened 2021-09-03
import earthstat data

elpamart:
I have included a logic for downloading the 175 crops from earthstat in Francis code. I've just included some parts for converting the nodata to -9999 value. We can change this if we want .

We will need to check the import in supervizz as this is super heavy - let me know what you think. My idea is to incorporate in this code the download process for the other data needed in the platform.


data/h3_raster_importer/tiff_folder_to_h3_table.py
```diff
@@ -89,6 +89,7 @@ def gen_raster_h3(raster_list, h3_res, geo=False):
                     raise ValueError("Transforms do not match")
                 arr = src.read(1, window=window)
                 _df = h3ronpy.raster.raster_to_dataframe(arr, w_transform, h3_res, nodata_value=src.profile['nodata'], compacted=False, geo=geo)
+                _df = _df.fillna(-9999)
```
fgassert, line 92: Maybe this value should be made a script parameter?

data/h3_raster_importer/Makefile
```diff
@@ -12,21 +12,42 @@ data/mapspam/spam2010v2r0_global_prod.geotiff.zip:
...
 	unzip -u $< *_A.tif -d $@
 
+
+data/earthstat/earthstat2000_global_prod: data/earthstat/HarvestedAreaYield175Crops_Geotiff.zip
+	mkdir -p $@
+	unzip -j -o -u $< *_Production.tif -d data/earthstat/earthstat2000_global_prod
+	for tiffile in data/earthstat/earthstat2000_global_prod/*.tif; do \
+		table_name=`basename -s .tif "$$tiffile" | tr -d ' \t\n\r' | tr [:upper:] [:lower:]`; \
+		output_file="data/earthstat/earthstat2000_global_prod/$${table_name}_noData.tif";\
+		gdal_translate -a_nodata -9999 -of GTiff "$${tiffile}" "$${output_file}";\
```
fgassert, line 35: What's the reason why you're setting the no-data value to -9999? Is it already encoded as -9999 in the GTiff, but not marked as no-data?

fgassert (commented):
Looks good! Just a question about how the no-data works

data/h3_raster_importer/tiff_folder_to_h3_table.py
fgassert: Also why are you filling the nodata back with -9999? Couldn't these be left to go into the DB as `null`? (or maybe even zero would be correct for the crop data)

data/h3_raster_importer/Makefile
elpamart: en mapspam the no data is marked as -1 and in earthstat the no data is marked as NA which means that we are ending up with a NA in an hexagon cell and I'm having issues for importing the table into the database.  I set that no data to -9999 randomly, and for testing that the import worked by setting any value to the no data, but its true that I can change it to -1 to have a similar config as mapspam. What do you think?

data/h3_raster_importer/tiff_folder_to_h3_table.py
elpamart: I'm setting the no data to -9999  for the same reason explained below - I did the same in this part of the code because (I don't know the reason why) I end up with NA values from the raster to h3 transformation even though I have already set the values to -9999 in the main script. I'll have another look today.

I didn't set this value to 0 to differentiate the no data from the areas with actual 0 value. I can set it to null or -1 as we have in mapspam.

fgassert: Hmm, ok I think it would be best to save them as nodata in the DB, but if they're not importing to the DB, maybe we need to configure the csv export like this: https://stackoverflow.com/questions/50890989/pandas-changing-the-format-of-nan-values-when-saving-to-csv?

data/h3_raster_importer/Makefile
fgassert: Actually it looks like EarthStat just has zeros, no nodata areas. Unless I have a different version?

Screen Shot 2021-09-07 at 9 36 42 AM

data/h3_raster_importer/tiff_folder_to_h3_table.py
elpamart: let me have a look

data/h3_raster_importer/Makefile
elpamart: ok, you are right. I was getting confused with the qgis info. I have set the no data in earthstat as 0, if that's ok?

fgassert: Yes I think that's good!

data/h3_raster_importer/tiff_folder_to_h3_table.py
elpamart: ok, changed the sql for importing the data to accept no data as 'NULL'.

alexeh:
Thanks @elpamart for pulling this out.
Just tested it locally and waterfootprint h3 table data is created / populated, and its info persisted in the H3 master table.

Looks good, thanks!

alexeh (approved):

PR #89, opened 2021-09-10
Feature/map legend

Bluesmile82:
First example of the map legend with a mocked choropleth (h3). Designs are still TBD.
- Used and styled scaffold legend
- Some adjustments to the layout have to be made to fix the overflowing and position of the map
- Used a hero icon. The scaffold icons don't seem to be working. Maybe we can fix that later


davidsingal:
Part of the agreements with the designers was to use heroicons, we may have to remove no needed components.

PR #90, opened 2021-09-14
Feature: LANDGRIF-176: Calculate Risk-Map 

alexeh:
This PR adds:

Implementation to generate a Risk-Map given a Material and a Indicator:

Risk-Map module: Service

The heavy work is done in the H3 Repository, as we can't get advantage of the plugin via queryBuilder, I am trying to keep all these raw queries in this repository

There are many many uncertainties here yet from Data point of view, and most likely there will be breaking changes in the data-model, so let's push forward about getting functionalities done, and let's rely on our e2e testing for a painless refactor later on

Suggestions, Complains and Threats always welcome!

**UPDATE:**
After chating with the LG Dev team, I refactored this PR moving the Risk-Map generation endpoint into the H3 controller, therefore deleting the Risk-Map standalone controller

This should be keep being a separated module since it will be more different calculus depending on the selected indicator so we can isolate all kind of risk-map generation into it


api/src/modules/indicators/indicator.entity.ts
```diff
@@ -65,4 +73,16 @@ export class Indicator extends BaseEntity {
   )
   @ApiPropertyOptional()
   indicatorCoefficients: IndicatorCoefficient[];
+
+  // TODO: Check with data if this relation can be stablished / make sense
```
tiagojsag, line 77: stablished => established

tiagojsag (approved):

PR #95, opened 2021-09-14
Feature/services to hooks

Bluesmile82:
Moved user and scenarios hooks from plain functions calling useQuery to hooks.

I don't think it's needed to do this for the rest as they are only calling UseQuery without any transformation or parameter.
Let me know if they should be changed. 

I couldn't test the user function as it is not called yet. Maybe if we have some more time we could do unit tests for the hooks and services.


davidsingal:
For now, the queries related to user authentication are not needed. Also, this is not completed in the API. I think everything should be in the same way to keep the consistency because it could be confusing for new developers to understand the criteria.

PR #101, opened 2021-09-21
Refactor H3 Response shape

alexeh:
This PR adds:

Refactoring H3 Response shape from a single object with:

`{'H3Index': 'valueForIndex'}`
To:
```
[
  {h: 'h3index', v: 'Value'}
]
```
Also introduces refactor for h3-repository to use queryBuilder and delete leftover raw query calls

Tests updated accordingly

### DIFF
--- a/api/src/modules/h3-data/h3-data.controller.ts
+++ b/api/src/modules/h3-data/h3-data.controller.ts
@@ -34,7 +34,7 @@ export class H3DataController {
   async geth3ByIdAndResolution(
     @Query(ValidationPipe)
     queryParams: MaterialH3ByResolutionDto,
-  ): Promise<Array<H3IndexValueData>> {
+  ): Promise<H3IndexValueData[]> {
     const { materialId, resolution } = queryParams;
     return await this.h3DataService.getMaterialH3ByResolution(
       materialId,
--- a/api/src/modules/h3-data/h3-data.controller.ts
+++ b/api/src/modules/h3-data/h3-data.controller.ts
@@ -4,7 +4,6 @@ import { H3DataService } from 'modules/h3-data/h3-data.service';
 import { H3Data, H3IndexValueData } from 'modules/h3-data/h3-data.entity';
 import { MaterialH3ByResolutionDto } from 'modules/h3-data/dto/h3-by-resolution.dto';
 import { GetRiskMapDto } from 'modules/risk-map/dto/get-risk-map.dto';
-import { RiskMapResponseDTO } from 'modules/risk-map/dto/response-risk-map.dto';
 import { RiskMapService } from 'modules/risk-map/risk-map.service';
 
 @Controller('api/v1/h3')
@@ -48,7 +47,7 @@ export class H3DataController {
   @Get('risk-map')
   async getRiskMap(
     @Query(ValidationPipe) queryParams: GetRiskMapDto,
-  ): Promise<RiskMapResponseDTO> {
+  ): Promise<H3IndexValueData[]> {
     const { materialId, indicatorId } = queryParams;
     return await this.riskMapService.calculateRiskMapByMaterialAndIndicator(
       materialId,
--- a/api/src/modules/h3-data/h3-data.controller.ts
+++ b/api/src/modules/h3-data/h3-data.controller.ts
@@ -20,7 +20,7 @@ export class H3DataController {
   async findOneByName(
     @Param('h3TableName') h3TableName: string,
     @Param('h3ColumnName') h3ColumnName: string,
-  ): Promise<{ data: H3IndexValueData }> {
+  ): Promise<{ data: H3IndexValueData[] }> {
     const h3Data = await this.h3DataService.findH3ByName(
       h3TableName,
       h3ColumnName,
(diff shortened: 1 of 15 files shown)


api/src/modules/h3-data/h3-data.controller.ts
```diff
@@ -34,7 +34,7 @@ export class H3DataController {
   async geth3ByIdAndResolution(
     @Query(ValidationPipe)
     queryParams: MaterialH3ByResolutionDto,
-  ): Promise<H3IndexValueData> {
+  ): Promise<Array<H3IndexValueData>> {
```
tiagojsag, line 37: I think you can also declare this as Promise<H3IndexValueData[]> 
This is FYI , no need to change it

api/src/modules/h3-data/h3-data.entity.ts
```diff
@@ -7,7 +7,10 @@ import {
 } from 'typeorm';
 
 export interface H3IndexValueData {
-  [key: string]: number;
+  //H3 index
+  h: string;
```
tiagojsag, line 11: triggered

api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -34,16 +38,9 @@ export class H3DataRepository extends Repository<H3Data> {
...
-   *
-   * @debt: This should be done on a single raw sql query, but I couldn't make it work
-   * Super Senior powers requested
    */
 
-  async getH3ByIdAndResolution(
-    h3Id: string,
-    resolution: number,
-  ): Promise<H3IndexValueData> {
+  async getH3ByIdAndResolution(h3Id: string, resolution: number): Promise<any> {
```
tiagojsag, line 43: Can the return type be improved?

api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -83,19 +72,13 @@ export class H3DataRepository extends Repository<H3Data> {
...
-     * @note Using query in order to parse the resultant array into a single hash-map
-     * so the frontend has a constant time complexity when manipulating this data
-     * also, wee need to attach + info about the map: measure unit in which the values should be shown, indicator data, etc..
-     * We could not do this using streams
-     */
+  ): Promise<H3IndexValueData[]> {
     const riskmap = await getManager()
       .createQueryBuilder()
-      .select('materialh3.h3index')
+      .select('materialh3.h3index', 'h')
```
tiagojsag, line 78: again...

triggered

if we are keeping the names short so we can shrink the FE response, please add a note saying so

api/test/h3-data/h3-data.spec.ts
```diff
@@ -104,7 +104,7 @@ describe('H3-Data Module (e2e)', () => {
       );
     });
 
-    test('When I query H3 data at minimal resolution, then I should get 8 h3indexes', async () => {
+    test.only('When I query H3 data at minimal resolution, then I should get 8 h3indexes', async () => {
```
tiagojsag, line 107: why the "only()" ?

tiagojsag (changes requested):

api/src/modules/h3-data/h3-data.controller.ts
alexeh: Totally correct, thanks for spotting it out. Will change it to keep code consistency and avoid my future self haunting me from the future.

Cheers!

api/src/modules/h3-data/h3-data.entity.ts
alexeh: - https://github.com/Vizzuality/landgriffon/pull/101/files#diff-46aa1ad8bbc9bdbf2afdf5de5ca7928aedf234a1881a252f83e6914c3d6b06dfR5

Sorry I though the interface definition comment was enough to clarify the purpose of the properties. A more extensive explanation available in the link above.

Thanks!

api/src/modules/h3-data/h3-data.repository.ts
alexeh: https://github.com/Vizzuality/landgriffon/pull/101/files#diff-46aa1ad8bbc9bdbf2afdf5de5ca7928aedf234a1881a252f83e6914c3d6b06dfR5

It was added actualy sir 💇

api/test/h3-data/h3-data.spec.ts
alexeh: Leftover of refactoring test, good catch, thanks!

api/src/modules/h3-data/h3-data.repository.ts
alexeh: Yes of course, good catch!

Fix -> 0752f2ebaedcc5d72f18086f93ed6623ebc71e32

api/src/modules/h3-data/h3-data.entity.ts
alexeh: Added **also** a comment to clarify the purspose of the interface and the reason behind the short prop names

Fix -> 0752f2ebaedcc5d72f18086f93ed6623ebc71e32

PR #107, opened 2021-09-29
Chore: LANDGRIF-270 Refactor Data-Model

alexeh:
Update relations
Fix data-import
Add risk-map by resolution
Simplify h3 generation flow and delete Risk-map module


api/src/modules/materials/material.entity.ts
```diff
@@ -112,25 +112,28 @@ export class Material extends TimestampedBaseEntity {
...
-  /**
-   * @note: Although there should not be any material with no h3GridId, this needs to be nullable
-   * since materials needs to exist before the h3 script takes care of adding these IDs to
-   * each material
-   */
+  @JoinColumn()
+  harvest: H3Data;
   @Column({ nullable: true })
-  h3GridId!: string;
+  harvestId: string;
```
tiagojsag, line 138: meta question: why "harvest"? what if the material is a mineral or something else?

api/src/modules/materials/materials.controller.ts
```diff
@@ -32,7 +32,7 @@ import { CreateMaterialDto } from 'modules/materials/dto/create.material.dto';
 import { UpdateMaterialDto } from 'modules/materials/dto/update.material.dto';
 import { MaterialRepository } from 'modules/materials/material.repository';
 import { ApiOkTreeResponse } from 'decorators/api-tree-response.decorator';
-import { ParseOptionalIntPipe } from 'pipes/parse-optional-int.pipe';
+import { ParseOptionalIntPipe } from 'pipes/parse-optional-int.pipe';import { Unit } from 'modules/units/unit.entity';
```
tiagojsag, line 35: wtf

api/test/risk-map/risk-map.spec.ts
```diff
@@ -107,62 +110,65 @@ describe('Risk Map Test Suite (e2e)', () => {
...
-      h3GridId: indicatorH3GridId,
     });
+    const indicatorH3Data = await createFakeH3Data(
+      INDICATOR_FAKE_H3_TABLE,
+      INDICATOR_FAKE_H3_COLUMN,
+      h3IndicatorFixtures,
+      indicator.id,
+    );
 
     const response = await request(app.getHttpServer()).get(
```
tiagojsag, line 170: use .query() instead if possible, makes code easier to maintain

tiagojsag (changes requested):
a few minor things here and there, overall LGTM once those are taken care of

api/src/modules/materials/material.entity.ts
alexeh: meta answer: naming does not reference the 'kind' of the material, but the type of the dataset, which we use then to determine which kind of calculus we need to do:
-Material Map -> producer
-Risk-Map -> harvest

Also, it is required by data to link existing materials to magically imported h3 without the API being involved

api/src/modules/materials/materials.controller.ts
alexeh: **alex grabbed his backpack and never returned. He was seen living naked in the mountains**

PR #123, opened 2021-10-19
Impact map endpoint

tiagojsag:


api/src/modules/import-data/sourcing-data/import.service.ts
```diff
@@ -109,6 +113,21 @@ export class SourcingRecordsImportService {
...
+
+      await Promise.all(
+        sourcingRecords.map(async (sourcingRecord: SourcingRecord) => {
+          await this.indicatorRecordsService.calculateImpactValue(
+            sourcingRecord,
+          );
+        }),
+      );
+
+      this.logger.log('Indicator records generated');
```
fgassert, line 130: Yes this looks generally right. Calculate the Impact Records async for all new Sourcing Records created.

fgassert: Theoretically this could/should all be wrapped in a transaction?

api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -45,12 +53,67 @@ export class IndicatorRecordsService extends AppBaseService<
...
+      | IndicatorRecord
+      | undefined = await this.indicatorRecordRepository.findOne(id);
     if (!found) {
       throw new NotFoundException(`Indicator Record with ID "${id}" not found`);
     }
 
     return found;
   }
+
+  async calculateImpactValue(
```
fgassert, line 68: Yes, looks good

api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -45,12 +53,67 @@ export class IndicatorRecordsService extends AppBaseService<
...
+            sourcingRecordId: sourcingRecord.id,
+          });
+        },
+      ),
+    );
+
+    this.logger.log(
+      `Impact value for sourcing record ${sourcingRecord.id} saved.`,
+    );
+    return this.indicatorRecordRepository.save(indicatorRecords);
```
fgassert, line 113: Just FYI for future optimizations; typeorm `save` is inefficient, as it does two queries first checking if the record exists then either insert or update. If you know which records to update, better to do a bulk delete, then bulk insert. Similarly, some queries could probably be batched in chunks.

api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -81,6 +94,27 @@ export class H3DataRepository extends Repository<H3Data> {
...
+    h3ColumnName: string,
+    geoRegionId: string,
+  ): Promise<number> {
+    // TODO: escape h3TableName and h3ColumnName
+    // TODO: try to use queryBuilder
+    const sum: { sum: number }[] = await getManager().query(
+      `
+        select sum(h3table.${h3ColumnName})
+        from ${h3TableName} h3table
+               left join (
```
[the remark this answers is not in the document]
fgassert: Note for future optimization: this is a great query to cache as the input params will not change, and for countries etc. you might have a lot of the same queries in a row.

api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -112,7 +146,96 @@ export class H3DataRepository extends Repository<H3Data> {
   }
 
   /**
-   *
+   * @param indicatorH3Data H3 format data of a Indicator
+   * @param calculusFactor Integer value to perform calculus. Represents the factor
+   * @param resolution Integer value that represent the resolution which the h3 response should be calculated
+   * @param groupBy
+   */
+  async getWaterImpactMapByResolution(
```
fgassert, line 154: I did not review this query/function

fgassert (approved):
Looks good to me!

PR #144, opened 2021-11-03
Filter out materials that have no h3 data

tiagojsag:


alexeh (approved):
Works like a charm. Thanks!

PR #147, opened 2021-11-04
Allow filtering by table when importing base data

tiagojsag:


elpamart (approved):
this is great! thanks :)

PR #167, opened 2021-11-18
Update API stack dependencies

tiagojsag:


api/config/default.json
```diff
@@ -18,7 +22,8 @@
     "database": "landgriffon",
     "synchronize": false,
     "migrationsRun": true,
-    "dropSchema": false
+    "dropSchema": false,
+    "logging": true
```
alexeh, line 26: So even tho I am now a proud #teamnative member, should we add an option into the dotenv skeleton file to make this option dynamic (without changing the code) for the poor people that haven't seen the light yet?

api/package.json
```diff
@@ -22,15 +22,16 @@
...
+    "@googlemaps/google-maps-services-js": "^3.3.2",
+    "@nestjs/axios": "^0.0.3",
+    "@nestjs/common": "^8.2.1",
+    "@nestjs/core": "^8.2.1",
+    "@nestjs/jwt": "^8.0.0",
+    "@nestjs/passport": "^8.0.1",
+    "@nestjs/platform-express": "^8.2.1",
+    "@nestjs/swagger": "^5.1.4",
+    "@nestjs/terminus": "^8.0.1",
+    "@nestjs/typeorm": "^8.0.2",
```
alexeh, line 34: I know this is the package you enjoyed the most updating :)
But shouldn't we aim to pin dependencies (at least the ones that are not development dependencies) ? 
Or there is a reason to not do so?

api/config/default.json
tiagojsag: https://github.com/typeorm/typeorm/blob/master/docs/logging.md

Typeorm expects a bool or an array here. I don't know how to put an array in a .env file, and I don't really feel like looking into that right now. If this is a concern, I can remove this change altogether, and we can tackle it later on, as part of a different task

api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -110,83 +109,82 @@ export class IndicatorRecordsService extends AppBaseService<
...
+        }
 
-          let impactValue: number | null = null;
+        let impactValue: number | null = null;
 
-          switch (indicator.nameCode) {
-            case 'UWU_T':
-              impactValue = await this.h3DataService.getWaterRiskIndicatorRecordValue(
+        switch (indicator.nameCode) {
+          case 'UWU_T':
```
alexeh, line 125: Hack alert: I suggest to use the indicator type enum both for readability and to avoid potential bugs due to magic string.
Not actually related to this PR, but reminding myself of a clear example of things to improve while cleaning

api/package.json
tiagojsag: I guess it makes more sense to change it to ~ so we can have bugfix versions updates. I'll take care of it

api/src/modules/indicator-records/indicator-records.service.ts
tiagojsag: +1

api/test/authentication/auth.spec.ts
```diff
@@ -32,26 +32,24 @@ describe('AppController (e2e)', () => {
...
-        user.password = await bcrypt.hash(
-          E2E_CONFIG.users.basic.aa.password,
-          user.salt,
-        );
-        user.isActive = true;
-        return user.save();
-      },
-    );
+    beforeAll(async (): Promise<User> => {
+      const user: User = new User();
```
alexeh, line 40: I think creating this mock user should not be placed in the beforeAll as it is not required for all the auth related use cases

alexeh (commented):
Thanks @tiagojsag for this PR. As it mixes stuff (update api dependencies, but also adds auth stuff) not sure if this question fits in this PR, but I failed to understand where do you avoid sending back salt, encrypted pass and other sensible stuff back to the client.

Could you please point me there?
I also added other minor questions related to the depedency changes

Thank you!

api/test/authentication/auth.spec.ts
tiagojsag: IIRC this is out of scope of this PR, it's just a code style change

tiagojsag:
I don't see where this PR adds auth stuff, it just reformats existing boilerplate code, because prettier was updated as part of the core motivation of this PR.

api/config/default.json
alexeh: I understand. Not really a concern to me, it is clear enough how to change between logging levels.
Let's tackle this if someone else complains about the logging levels (which most likely won't happen)

alexeh:
> I don't see where this PR adds auth stuff, it just reformats existing boilerplate code, because prettier was updated as part of the core motivation of this PR.

Yes you right, I probably got 'so confused that I injured myself' due to the amount of changes here, apologies for that.
So feel free to merge this PR once the green lights are shining, and we can address the auth related comments later on.

Thanks!

PR #168, opened 2021-11-19
Refactor sql queries to typeorm

yulia-bel:


api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -225,49 +233,59 @@ export class H3DataRepository extends Repository<H3Data> {
...
-              AND def.hansen_loss_2019 > 0
-              AND ind.${indicatorH3Table.h3columnName} > 0) materials --replace by ind column name
-           ON materials.h3index = locations.h3index
-      GROUP BY locations.tonnage
-    `;
-
-    const result: { value: number }[] = await getManager().query(query);
+    const result: { value: number }[] = await getManager()
+      .createQueryBuilder()
+      .select('locations.tonnage * sum(materials.carb_risk_CO2e_t)', 'value')
```
alexeh, line 238: Given we are already refactoring some of the hacks, maybe would a good idea to use more 'readable' names for the tables? i.e carbon_risk_CO2e_t concatenates the unit that should be shown with this data, which has nothing to do with the query itself I think.

api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -225,49 +233,59 @@ export class H3DataRepository extends Repository<H3Data> {
...
+      }, 'locations')
+      .innerJoin(
+        (subQuery: SelectQueryBuilder<any>) => {
+          return subQuery
+            .select(`ha.h3index`)
+            .addSelect(
+              `ind.${indicatorH3Table.h3columnName} *
+            ((def.hansen_loss_2019 * ha.${harvestH3Table.h3columnName}) / (sum(prod.${producerH3Table.h3columnName}) over ())) carb_risk_CO2e_t`,
+            )
+            .from(harvestH3Table.h3tableName.toLowerCase(), 'ha')
```
alexeh, line 263: Are these` .toLowerCase() `required? I think we retrieve these magical table names / column names, as they already exist in the BD, perform transformations within the API could cause troubles maybe?

api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -285,49 +303,60 @@ export class H3DataRepository extends Repository<H3Data> {
...
+          .addSelect('unnest(gr."h3Flat"::h3index[])', 'h3index')
+          .from('sourcing_records', 'sr')
+          .leftJoin(
+            'sourcing_location',
+            'sl',
+            'sl.id = sr."sourcingLocationId"',
+          )
+          .leftJoin('geo_region', 'gr', 'gr.id = sl."geoRegionId"')
+          .where(`sr.id = :sourcingRecordId`, {
+            sourcingRecordId: sourcingRecordId,
```
alexeh, line 322: could be this be shorthanded to 
```suggestion
{ sourcingRecordId }
```

alexeh (commented):
Thanks @yulia-bel! I left some over it, I'd like to know @tiagojsag point of view about them as well.

Nice work!

PR #177, opened 2021-11-26
Fix regressions following case change in data tables

tiagojsag:


api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -77,10 +77,10 @@ export class H3DataRepository extends Repository<H3Data> {
       const query: string = getManager()
         .createQueryBuilder()
         .select(`h3_to_parent(h3index, ${resolution})`, 'h')
-        .addSelect(`sum(${materialH3Data.h3columnName})`, 'v')
-        .from(`${materialH3Data.h3tableName}`, 'h3table')
-        .where(`${materialH3Data.h3columnName} is not null`)
-        .andWhere(`${materialH3Data.h3columnName} <> 0`)
+        .addSelect(`sum("${materialH3Data.h3columnName}")`, 'v')
+        .from(materialH3Data.h3tableName, 'h3table')
+        .where(`"h3table"."${materialH3Data.h3columnName}" is not null`)
```
alexeh, line 82: These double quotes are not needed as the alias is already in lowercase

alexeh (approved):
Thanks!

api/src/modules/h3-data/h3-data.repository.ts
tiagojsag: I know, but it also doesnt hurt, so...

PR #179, opened 2021-11-29
Refactor indicator types

yulia-bel:


api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -149,7 +149,7 @@ export class IndicatorRecordsService extends AppBaseService<
                 sourcingRecord.id,
               );
             break;
-          case 'GHG_LUC_T':
+          case 'carbon_emissions':
```
alexeh, line 152: Even tho now is way more readable, we are still using magic strings here. 

Please have a look to h3-data.service.ts, function `getRiskMapByResolution` to check how we are avoiding using said magic strings there to get an idea

api/test/import-data/xlsx-uploads/import-mocks.ts
```diff
@@ -8,19 +8,19 @@ async function createIndicatorsForXLSXImport(): Promise<string[]> {
...
       name: 'Biodiversity loss due to land use change',
-      nameCode: 'BL_LUC_T',
+      nameCode: 'biodiversity_loss',
     },
     {
       name: 'Unsustainable water use',
-      nameCode: 'UWU_T',
+      nameCode: 'unsustainable_water_use',
     },
   ];
```
alexeh, line 25: You can also use the LAYER_TYPE enum for the tests

alexeh (changes requested):
Thanks @yulia-bel!, I requested some changes below 😃

api/test/import-data/xlsx-uploads/import-mocks.ts
alexeh: Hm weird, the issue does not appear as outdated even tho I see you made the suggested change 🤔

alexeh (approved):
LGTM, thanks @yulia-bel !

PR #180, opened 2021-11-29
LANDGRIF-368 - Implement optional filters for impact map endpoint

tiagojsag:

PR #182, opened 2021-11-30
Landgrif-380 - Remove groupby from impact map

tiagojsag:
Merge & rebase https://github.com/Vizzuality/landgriffon/pull/180 first


alexeh (approved):
LGTM, thank you!

davidsingal (changes requested):
We are using the same component for all the filters for map, chart, and table. That's why we cannot remove it easily.

davidsingal (approved):

PR #184, opened 2021-12-02
Add material map test based on alternative fixture

yulia-bel:


api/test/h3-data/h3-material-map.spec.ts
```diff
@@ -106,4 +111,31 @@ describe('H3 Data Module (e2e) - Material map', () => {
...
+
+  test('When I query H3 data at minimal resolution, then I should get 4 h3indexes with expected values and no 0 as value', async () => {
+    const h3Data = await h3AlternativeDataMock(
+      fakeTable,
+      fakeColumn,
+      h3AlternativeFixture,
+    );
+    const material = await createMaterial({ producer: h3Data });
+    const response = await request(app.getHttpServer()).get(
+      `/api/v1/h3/map/material?materialId=${material.id}&resolution=1`,
```
tiagojsag, line 123: look into the .query() to add query arguments, instead of adding them to the URL string.

alexeh: +1

api/test/h3-data/mocks/h3-data.mock.ts
```diff
@@ -39,6 +39,43 @@ export const h3DataMock = async (
...
+  h3data.h3columnName = formattedColumnName;
+  h3data.h3resolution = 6;
+  h3data.year = year;
+  if (indicatorId) {
+    h3data.indicatorId = indicatorId;
+  }
+
+  return h3data.save();
+};
+
```
alexeh, line 78: Since you added a more accurate fixture for a fake material h3 data, maybe we can just use that fixture on the main h3DataMock function (or the one you just created) instead of having both, which seems to do the same but with different fixtures?

alexeh (approved):
LGTM Thanks!

api/test/h3-data/h3-risk-map.spec.ts
```diff
@@ -34,8 +55,10 @@ describe('H3 Data Module (e2e) - Risk map', () => {
   let unitConversionRepository: UnitConversionRepository;
   const fakeTable = 'faketable';
```
tiagojsag, line 56: question for you and @alexeh : what value do these add? I don't really see a reason to have these constants, and I think code readability would improve if we just remove them

api/test/h3-data/h3-risk-map.spec.ts
```diff
@@ -170,44 +193,113 @@ describe('H3 Data Module (e2e) - Risk map', () => {
...
+    const responseRes3 = await request(app.getHttpServer())
+      .get(`/api/v1/h3/map/risk`)
+      .query({
+        indicatorId: indicator.id,
+        year: 2020,
+        materialId: material.id,
+        resolution: 3,
+      });
+
+    expect(h3DataRepository.getWaterRiskMapByResolution).toHaveBeenCalledTimes(
```
tiagojsag, line 284: IMO this sort of spy usage is redundant - if, for some reason, this method was not called, the result would differ, and your test would pick it up anyway. Plus, it ties your test to the implementation (if you implement the exact same feature, with a different method, your test would yield a false negative)

@alexeh I know you are a fan, so maybe you'd like to offer a different take on this?

tiagojsag: also: if we decide to keep it, we should validate the arguments

api/test/h3-data/h3-risk-map.spec.ts
```diff
@@ -344,6 +509,13 @@ describe('H3 Data Module (e2e) - Risk map', () => {
       fakeColumn,
     });
 
+    const indicatorH3Data = await h3DataMock(
```
tiagojsag, line 518: you can remove the `const indicatorH3Data = ` part, as it's never referenced again (your IDE should probably pick this up for you)

api/test/h3-data/h3-risk-map.spec.ts
```diff
@@ -368,6 +540,13 @@ describe('H3 Data Module (e2e) - Risk map', () => {
...
```
tiagojsag, line 550: same here as above

api/test/h3-data/h3-risk-map.spec.ts
```diff
@@ -269,51 +386,83 @@ describe('H3 Data Module (e2e) - Risk map', () => {
...
+      expectedCarbonEmissionsRes6Metadata,
+    );
+    expect(responseRes3.body.data).toEqual(
+      expect.arrayContaining(expectedCarbonEmissionsRes3Data),
+    );
+    expect(responseRes3.body.metadata).toEqual(
+      expectedCarbonEmissionsRes3Metadata,
+    );
   });
   // TODO: Update assertion as soon as actual calculus is validated
```
tiagojsag, line 455: @yulia-bel @alexeh is this still relevant?

api/test/h3-data/mocks/h3-alternative-fixture.ts
```diff
@@ -1,5 +1,7 @@
 export const h3AlternativeFixture = {
```
tiagojsag, line 1: Overall I don't like the "alternative" name used all over the place. If someone joined the project tomorrow, and saw this, they would think "ok, they are alternative, but alternative to what? Did we have some other fixtures before, or are these just hipster fixtures that listen to music that hasn't been made yet?"

Additionally, since they are hard to decode mentally, I would suggest adding a small comment on each of them, describing what they are useful for. Again, thinking about the "how could this help a dev that joins the project tomorrow" usually helps in writing a good description.

tiagojsag: same thing on the "expected-responses" part

tiagojsag (changes requested):
Overall great stuff, just a bunch of minor things in the comments. Perhaps the biggest concert I have with this PR is that it seems to add fixtures/mock data, instead of replacing it. We don't necessarily need to remove/replace things in this PR, but we should at least have a slightly broader plan to either refactor or replace old fixtures, so we end up with something that is easy to understand. Right now we have the old stuff, and the "alternative" stuff, and it's not clear why we have both, or what's used where.

api/test/h3-data/h3-risk-map.spec.ts
alexeh: Good point @tiagojsag . Since this aims to replace a magic table, the name does not really matter (IMO) and its at the main describe block scope to be reused in more than one tests, instead of adding a new string a instantiate a new constant in each test block. 

I wouldn't mind if you consider removing it and then use a new string when creating a fake h3 table is required

api/test/h3-data/h3-risk-map.spec.ts
alexeh: contrary to what everyone might think, I do agree with @tiagojsag . So I implemented these spies in the risk-map test suite in the time that some calculations weren't different (they have been changing in time) so it was, in my opinion, a good idea to know that depending on requested type, X service was called. Also, my understanding of H3 stuff was (even more) low

Now that we have a more clear understanding, and clearly different calcs for each kind of risk-map, I do agree that this could be removed, or, validate the arguments that it has been called

api/test/h3-data/h3-risk-map.spec.ts
alexeh: I guess **NOW** it's the moment to remove it! :)

api/test/h3-data/mocks/h3-alternative-fixture.ts
alexeh: I second @tiagojsag thought on replacing the old fixtures, as they make no longer sense thanks to your brilliant contribution.

IMO the expected-responses part does no harm and it's purpose is quite understandable, but probably @tiagojsag can see further than we actually can

yulia-bel: Thank you for the reviews @tiagojsag and @alexeh! I think all the comments have been addressed - old fixtures replaced by the new ones, no more alternatives :) I also changed the naming a bit and added short descriptions in the fixture files.

PR #185, opened 2021-12-02
LANDGRIF-373 - Refactor h3Data - material relationship

tiagojsag:


api/src/modules/h3-data/dto/get-material-h3-by-resolution.dto.ts
```diff
@@ -27,4 +27,10 @@ export class GetMaterialH3ByResolutionDto {
   @IsNumber()
   @IsEnum(AvailableResolutions, { message: 'Available resolutions: 1 to 6' })
   resolution!: number;
+
+  @ApiProperty()
+  @Type(() => Number)
+  @IsNumber()
+  @IsNotEmpty()
+  year!: number;
 }
```
alexeh, line 36: Right, this param was needed indeed and I should have added it myself. thanks!

api/src/modules/materials/materials.service.ts
```diff
@@ -60,7 +60,7 @@ export class MaterialsService extends AppBaseService<
    */
   static filterMaterialTree(root: Material): Material[] {
     let result: Material[] = [];
-    if (root.producerId === null && root.harvestId === null) {
+    if (!root.materialToH3s || root.materialToH3s.length === 0) {
```
alexeh, line 63: I think this could also be shorthanded to  !root.materialToH3s.length

yulia-bel (approved):

tiagojsag:
Thanks for the feedback, I think I got all of your comments addressed, please let me know if I missed anything

alexeh (approved):
LGTM, Thanks for this BIG task!

PR #186, opened 2021-12-09
improve in existing raw materials

elpamart:
This PR contains changes in the materials csv and in the generation of aggregates materials groups. For example, the parent material cereals is now generated by aggregating all the raw raster from their childrens. This aggregation is calculated in gdal with the data import script, translated to h3 and ingested/linked to the materials entity.


alexeh (approved):
Since I can't really test this PR without the risk of my Vizz laptop exploding, can you please @tiagojsag spin this up, so we can check everything is fine and merge it to staging?

PR #188, opened 2021-12-13
Refactor sql to typeorm

yulia-bel:
Since typeorm doesn't support WITH clause and INSERT INTO .. SELECT the first query refactor is very limited, maybe it doesn't make sense to refactor it at all..

### DIFF
--- a/api/src/modules/geo-regions/geo-region.repository.ts
+++ b/api/src/modules/geo-regions/geo-region.repository.ts
@@ -1,4 +1,9 @@
-import { EntityRepository, Repository } from 'typeorm';
+import {
+  EntityRepository,
+  getManager,
+  Repository,
+  SelectQueryBuilder,
+} from 'typeorm';
 import { GeoRegion } from 'modules/geo-regions/geo-region.entity';
 import { LocationGeoRegionDto } from 'modules/geo-regions/dto/location.geo-region.dto';
 
@@ -20,18 +25,24 @@ export class GeoRegionRepository extends Repository<GeoRegion> {
   async saveGeoRegionAsRadius(
     newGeoRegionValues: LocationGeoRegionDto,
   ): Promise<GeoRegion> {
+    const selectQuery: any = getManager()
+      .createQueryBuilder()
+      .select(`hashtext($3)`)
+      .addSelect(`points.radius`)
+      .addSelect(
+        `array(
+        SELECT h3_compact(array(SELECT h3_polyfill(points.radius,6)))
+      )`,
+      )
+      .from('points', 'points');
+
     const res: any = await this.query(
-      `WITH
+      `
+      WITH
         points AS (SELECT ST_BUFFER(ST_SetSRID(ST_POINT($1,$2),4326)::geometry, 0.5) as radius)
       INSERT INTO geo_region (name, "theGeom", "h3Compact")
-      SELECT
-        hashtext($3) ,
-        points.radius,
-        array(
-          SELECT h3_compact(array(SELECT h3_polyfill(points.radius,6)))
-        )
-      FROM points
-        ON CONFLICT (name) DO UPDATE
+      ${selectQuery.getSql()}
+      ON CONFLICT (name) DO UPDATE
           SET "theGeom" = excluded."theGeom", "h3Compact" = excluded."h3Compact"
         RETURNING id`,
       [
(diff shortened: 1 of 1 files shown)


api/src/modules/geo-regions/geo-region.repository.ts
```diff
@@ -20,18 +20,23 @@ export class GeoRegionRepository extends Repository<GeoRegion> {
...
+        `array(
+        SELECT h3_compact(array(SELECT h3_polyfill(points.radius,6)))
+      )`,
+      )
+      .from('points', 'points');
+
     const res: any = await this.query(
       `WITH
         points AS (SELECT ST_BUFFER(ST_SetSRID(ST_POINT($1,$2),4326)::geometry, 0.5) as radius)
       INSERT INTO geo_region (name, "theGeom", "h3Compact")
```
tiagojsag, line 37: this part can probably also go in a querybuilder syntax (so the selectQuery would actually be an insertQuery

api/src/modules/geo-regions/geo-region.repository.ts
```diff
@@ -52,23 +57,30 @@ export class GeoRegionRepository extends Repository<GeoRegion> {
...
-      [
+    const res: any = await getManager()
+      .createQueryBuilder()
+      .insert()
+      .into('geo_region')
+      .values({
+        name: () => `hashtext(:arg1)`,
+
+        theGeom: () => `ST_GeomFromText(:arg2, 4326)`,
+
```
tiagojsag, line 68: empty lines here? does prettier allow this?

tiagojsag (changes requested):

api/src/modules/geo-regions/geo-region.repository.ts
yulia-bel: The problem is that it seems that typeorm doesn't support INSERT INTO .. SELECT (https://github.com/typeorm/typeorm/issues/4374), and every time insert() is used, it requires values(). 
I tried 3 different approaches, none worked:
1. Add insert() part to queryBuilder followed by select() instead of values, - the insert part is ignored, translated query starts with select:
```const insertQuery: any = getManager()
      .createQueryBuilder()
      .insert()
      .into('geo_region')
      .select(`hashtext($3)`)
      .addSelect(`points.radius`)
      .addSelect(
        `array(
        SELECT h3_compact(array(SELECT h3_polyfill(points.radius,6)))
      )`,
      )
      .from('points', 'points'); 
```
SQL result:
``` 
WITH points AS (SELECT ST_BUFFER(ST_SetSRID(ST_POINT($1,$2),4326)::geometry, 0.5) as radius)
            SELECT hashtext($3), points.radius, array(
              SELECT h3_compact(array(SELECT h3_polyfill(points.radius,6)))
            ) FROM "geo_region" "geo_region", "points" "points"
            ON CONFLICT (name) DO UPDATE
                SET "theGeom" = excluded."theGeom", "h3Compact" = excluded."h3Compact"
              RETURNING id -- PARAMETERS: [78.96288,20.593684,"78.96288-20.593684"] 
```
2. Combine 2 queries (insert and select) written with typeorm and insert them in raw query with getSql():
```
const insertQuery: any = getManager()
      .createQueryBuilder()
      .insert()
      .into('geo_region');

const selectQuery: any = getManager()
      .createQueryBuilder()
      .select(`hashtext($3)`)
      .addSelect(`points.radius`)
      .addSelect(
        `array(
        SELECT h3_compact(array(SELECT h3_polyfill(points.radius,6)))
      )`,
      )
      .from('points', 'points');

const res: any = await this.query(
      `WITH
        points AS (SELECT ST_BUFFER(ST_SetSRID(ST_POINT($1,$2),4326)::geometry, 0.5) as radius)
      ${insertQuery.getSql()} 
      ${selectQuery.getSql()}
      ON CONFLICT (name) DO UPDATE
          SET "theGeom" = excluded."theGeom", "h3Compact" = excluded."h3Compact"
        RETURNING id`
```
It doesn't work, because typeorm requires .values() after insert():
```
InsertValuesMissingError: Cannot perform insert query because values are not defined. Call "qb.values(...)" method to specify inserted values.
```

3. The last approach I tried was adding values() to the queryBuilder and putting the select inside values() in different ways. Below is the furthest I could get and it is giving undefined values where 'points' are involved. (without subqueries from() couldn't be added after values())
Plus the sql query is different - it is INSERT() INTO .. VALUES() -  not INSERT() INTO ... SELECT(), so it is not just translating to typeorm, it is refactoring the query as well, so I thought it is not the best way. But if you think it could be the right direction - I will keep looking deeper into it
```
    const insertQuery: any = getManager()
      .createQueryBuilder()
      .insert()
      .into('geo_region')
      .values({
        name: () => `hashtext($3)`,
        theGeom: () => {
          (subQuery: SelectQueryBuilder<any>) => {
            return subQuery.select(`points.radius`).from(`points`, `points`);
          };
        },
        h3Compact: () => {
          (subQuery: SelectQueryBuilder<any>) => {
            return subQuery
              .select(
                `array(
              SELECT h3_compact(array(SELECT h3_polyfill(points.radius,6)))
            )`,
              )
              .from(`points`, `points`);
          };
        },
      })
      .orUpdate(['theGeom', 'h3Compact'], ['name'])
      .returning('id');
```

tiagojsag: thorough explanation <3

api/src/modules/geo-regions/geo-region.repository.ts
```diff
@@ -20,18 +20,23 @@ export class GeoRegionRepository extends Repository<GeoRegion> {
   async saveGeoRegionAsRadius(
     newGeoRegionValues: LocationGeoRegionDto,
   ): Promise<GeoRegion> {
+    const selectQuery: any = getManager()
```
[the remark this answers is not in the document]
tiagojsag: +1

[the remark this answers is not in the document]
yulia-bel: This particular selectQuery is selecting not from the GeoRegion entity, but from 'points' created by WITH clause. 
However, I replaced getMangaer() of the second insert query by extended class Query Builder as you mentioned - thanks
Also added SelectQueryBuilder and InsertResult types

tiagojsag (approved):

alexeh (approved):
LGTM, thanks @yulia-bel

PR #194, opened 2021-12-15
Resolution 6 when a origin is selected only for impact layer

davidsingal:


davidsingal:
I've tried with res 9, but it seems like the API has restricted max resolution.

elpamart (approved):

PR #197, opened 2021-12-20
Refactor materialToH3 typeorm query

yulia-bel:


alexeh (approved):

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -165,7 +165,7 @@ export class H3DataService {
      */
     const indicatorH3Data: H3Data | undefined =
       await this.h3DataRepository.findOne({
-        where: { indicatorId: indicatorId },
+        where: { indicatorId: indicatorId, year: year },
```
tiagojsag, line 168: the message on line 184 should be updated

tiagojsag (changes requested):

api/src/modules/h3-data/h3-data.service.ts
yulia-bel: Thanks! Added year mentioning to the error message. This message will probably change again once we have fallback strategy implemented? Since if the material won't be found just for the year, some alternative will bee provided..

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -165,12 +165,12 @@ export class H3DataService {
...
       });
 
     if (!indicatorH3Data)
       throw new NotFoundException(
-        `There is no H3 Data for Indicator with ID: ${indicatorId}`,
+        `There is no H3 Data for Indicator with ID: ${indicatorId} for the requested year`,
```
tiagojsag, line 173:
```suggestion
        `There is no H3 Data for Indicator with ID ${indicatorId} and year ${year}`,
```

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -181,7 +181,7 @@ export class H3DataService {
       });
     if (!harvestMaterialH3Data) {
       throw new NotFoundException(
-        `There is no H3 harvest data for Material with ID: ${materialId}`,
+        `There is no H3 harvest data for Material with ID: ${materialId} for the requested year`,
```
tiagojsag, line 184:
```suggestion
        `There is no H3 harvest data for Material with ID ${materialId} and year ${year}`,
```

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -193,7 +193,7 @@ export class H3DataService {
       });
     if (!producerMaterialH3Data) {
       throw new NotFoundException(
-        `There is no H3 producer data for Material with ID: ${materialId}`,
+        `There is no H3 producer data for Material with ID: ${materialId} for the requested year`,
```
tiagojsag, line 196:
```suggestion
        `There is no H3 producer data for Material with ID ${materialId} and year ${year}`,
```

tiagojsag (changes requested):
review my suggestions, update the tests, and should be good to go. Having this sort of information is very useful for debugging, both to find the error, as well as to give us an idea of what data may be missing - if I had added this in my years PR, we probably would have found this bug a lot sooner and easier

PR #200, opened 2021-12-23
Update name hash when saving as radius

yulia-bel:


alexeh (changes requested):
Thanks @yulia-bel!

We should add a test / refactor a existing test (I'd go for the latter) to assert this new behaviour, and make the tests blow if someone accidentally change it

api/src/modules/geo-regions/geo-region.repository.ts
```diff
@@ -49,7 +48,7 @@ export class GeoRegionRepository extends Repository<GeoRegion> {
       [
         newGeoRegionValues.coordinates.lng,
         newGeoRegionValues.coordinates.lat,
-        `${newGeoRegionValues.coordinates.lng}-${newGeoRegionValues.coordinates.lat}`,
+        `concat('${newGeoRegionValues.coordinates.lng}-${newGeoRegionValues.coordinates.lat}', ', radius - ', points.radius)`,
```
tiagojsag, line 51: any reason why you prefer concat over https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals ?

yulia-bel: CONCAT() here is for postgres to put 'points.radius' into the hashtext(), since it is not a parameter but something  achieved by a select within the query (in WITH clause). It was not working how it was expected though, it is refactored now

tiagojsag (approved):

PR #201, opened 2021-12-31
Implement missing years fallback strategy for risk map

yulia-bel:


api/test/h3-data/h3-risk-map.spec.ts
```diff
@@ -279,27 +284,137 @@ describe('H3 Data Module (e2e) - Risk map', () => {
...
+
+    await h3DataMock({
+      h3TableName: 'fakeIndicatorTable',
+      h3ColumnName: 'fakeIndicatorColumn',
+      additionalH3Data: h3IndicatorExampleDataFixture,
+      indicatorId: indicator.id,
+      year: 2005,
+    });
+
+    jest.spyOn(h3DataRepository, 'getWaterRiskMapByResolution');
```
tiagojsag, line 308: what is this meant to do?

tiagojsag (changes requested):
only a minor thing (that happens a few times), otherwiselogic looks good:)

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -163,20 +168,49 @@ export class H3DataService {
...
      * @note To generate a Risk Map, harvest data and h3Data by indicatorId are required
      */
+
+    const availableIndicatorYears: number[] =
+      await this.getAvailableYearsForH3IndicatorData(indicatorId);
+
+    const indicatorDataYear: number | undefined =
+      availableIndicatorYears.includes(year)
+        ? year
+        : availableIndicatorYears.find((el) => el < year);
```
alexeh, line 178: What would happen here if we actually have data for the requested year? Wouldn't we performing calcs with for the previous year?

`const arr = [2004, 2003, 2002, 2001, 2000]

 const year = 2002

 const availableYear = arr.find((el) => el < year)

 console.log(availableYear) // 2001`

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -185,10 +219,26 @@ export class H3DataService {
...
+    const availableProducerDataYears: number[] =
+      await this.getAvailableYearsForH3MaterialData(
+        materialId,
+        MATERIAL_TO_H3_TYPE.PRODUCER,
+      );
+
+    const producerDataYear: number | undefined =
+      availableProducerDataYears.includes(year)
+        ? year
+        : availableProducerDataYears.find((el) => el < year);
```
alexeh, line 231: Same comment [as above](https://github.com/Vizzuality/landgriffon/pull/201/files#r777018934)

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -352,4 +408,22 @@ export class H3DataService {
...
+    );
+  }
+
+  async getAvailableYearsForH3IndicatorData(
+    indicatorId: string,
+  ): Promise<number[]> {
+    return await this.h3DataRepository.getAvailableYearsForH3IndicatorData(
+      indicatorId,
+    );
+  }
```
alexeh, line 428: Maybe both of these methods would exist on the 

`h3-filter-years-by-layer.service.ts` ? (maybe with better naming :))

As these are also related getting available years

api/src/modules/h3-data/dto/h3-map-response.dto.ts
```diff
@@ -33,9 +33,24 @@ export class H3MapResponse {
...
+      },
     },
   })
-  metadata: { unit: string; quantiles: number[] };
+  metadata: {
+    unit: string;
+    quantiles: number[];
+    indicatorDataYear?: number;
+    harvestDataYear?: number;
+    producerDataYear?: number;
```
alexeh, line 52: Even tho producer h3 data is required to do magic when it comes to risk-maps, it is just used internally to do said magic (and for some type of indicators) So I think we can safely remove it from the response

Also I think this data is included so the user can know with which years we sent a response, in case the we have no data for the requested year. So maybe avoid adding the value in case we have data for matching years? Or not worht the effort?

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -163,20 +168,49 @@ export class H3DataService {
...
+        : availableIndicatorYears.find((el) => el < year);
+
+    if (!indicatorDataYear)
+      throw new NotFoundException(
+        `There is no H3 Data registered for Indicator with ID ${indicatorId} for the year ${year} or any year before ${year}`,
+      );
     const indicatorH3Data: H3Data | undefined =
       await this.h3DataRepository.findOne({
-        where: { indicatorId: indicatorId, year: year },
+        where: { indicatorId: indicatorId, year: indicatorDataYear },
```
alexeh, line 186:
```suggestion
        where: { indicatorId, year: indicatorDataYear },
```

alexeh (commented):
Thanks @yulia-bel great work! I left some comments for clarification.

I am missing the test for when we have matching years, I think (I may be wrong) there is a little bug about this case as I stated in some comments.

Thank you!

api/test/h3-data/h3-risk-map.spec.ts
alexeh: Seems that the couple of spies present are not used in the assertions, I guess we can remove them :)

api/src/modules/h3-data/h3-data.service.ts
yulia-bel: thanks :)

api/src/modules/h3-data/dto/h3-map-response.dto.ts
yulia-bel: Sure, I think also we can ask Elena and David about how they prefer to inform user about the the years situation, for us it is a minor detail of what to put in response

api/src/modules/h3-data/h3-data.service.ts
yulia-bel: Yes, at first I used this service, but then due to failing test and code investigation I found out that I cannot use it - because: 
- h3-filter-years-by-layer is retrieving available years for the layer (for example, material)
- h3DataIds is an optional parameter of this service
- the service tries to find h3DataIds by the given material Id, if no h3DatIds are found - the server just doesn't include the filter by h3DataIds in the query and returns all the years available for the layer (material)  in general. In our case it is not correct, because we need available years for the particular material and its H3Data, if no h3Data is found, no years shall be returned and the error should be thrown (saying there is no h3 data for that material for that year or any other year in the past)

api/test/h3-data/h3-risk-map.spec.ts
yulia-bel: it was in the test before, I am not sure I understand 100% what exactly it does here, but I will investigate it

alexeh: I think [ this is the answer](https://github.com/Vizzuality/landgriffon/pull/184#discussion_r765837609) to that question

api/src/modules/h3-data/h3-data.service.ts
alexeh: Sorry. maybe my comment was not clear (probably)

What I meant is we could encapsulate your logics:
`getAvailableYearsForH3IndicatorData`
and
`getAvailableYearsForH3MaterialData`

into the standalone service which does stuff about years, so we can have all stuff related to years encapsulated in the same class (the class itself does not receive any params related to years, layer or anything, just its dependencies by injection)

However this is code stylish stuff, let's see the team's thoughts

api/test/h3-data/h3-risk-map.spec.ts
tiagojsag: ok, so let's remove them

api/src/modules/h3-data/h3-data.service.ts
tiagojsag: implementation detail, either way is fine, leave it as it is

alexeh (approved):

tiagojsag (approved):

PR #202, opened 2022-01-03
chore/LANDGRIF-390 Make excel import async

alexeh:
This PR adds: 

First piece of [Making stuff async](https://vizzuality.atlassian.net/jira/software/c/projects/LANDGRIF/boards/14?modal=detail&selectedIssue=LANDGRIF-236):

Add redis to compose setup
Implement a queue system for the import-module
Add some flags for testing increasing node heap size and some flags that will let us know memory usage on test from now on
Some renaming

NOTE:
Tracking job events not in this PR's scope
Sourcing-Data import tests are skipped as it will no longer make sense, not deleting as I'll refactor them on the next PR


.github/workflows/testing-api.yml
```diff
@@ -28,6 +29,12 @@ jobs:
           POSTGRES_USER: landgriffon
           POSTGRES_PASSWORD: landgriffon
           POSTGRES_DB: landgriffon
+          NODE_OPTIONS: --max-old-space-size=4096
```
tiagojsag, line 32: postgres doesnt need extra heap space (but it thanks you for it anyway)

.github/workflows/testing-api.yml
```diff
@@ -11,6 +11,7 @@ env:
   DB_USERNAME: landgriffon
   DB_PASSWORD: landgriffon
   DB_DATABASE: landgriffon
+  API_REDIS_SERVICE_HOST: localhost
```
tiagojsag, line 14: this is only used on the docker-compose file, which is not used on GH actions

api/package.json
```diff
@@ -13,10 +13,10 @@
     "start:prod": "NODE_OPTIONS=\"--max-old-space-size=2048\" NODE_PATH=dist/src node dist/src/main",
     "build": "rm -rf dist && tsc",
     "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
-    "test": "jest --config test/jest-config.json -i",
+    "test": "NODE_OPTIONS=\"--max-old-space-size=4096\" node --expose-gc ./node_modules/.bin/jest --config test/jest-config.json -i --logHeapUsage --detectOpenHandles --forceExit ",
```
tiagojsag, line 16: do we really need all of this?

api/test/jest-config.json
```diff
@@ -31,5 +31,10 @@
   "coveragePathIgnorePatterns": [
     "<rootDir>/node_modules/",
     "<rootDir>/test/"
-  ]
+  ],
+  "globals": {
+    "ts-jest": {
+      "isolatedModules": true
```
tiagojsag, line 37: what does this do, and why do we need it?

docker-compose.yml
```diff
@@ -28,15 +28,43 @@ services:
       - ./api/test:/opt/landgriffon-api/test
     env_file: .env
     environment:
+      - NODE_OPTIONS=${HEAP_SIZE}
```
tiagojsag, line 31: ???

docker-compose.yml
```diff
@@ -28,15 +28,43 @@ services:
...
       - DB_SYNCHRONIZE=${API_DB_SYNCHRONIZE}
       - JWT_SECRET=${API_AUTH_JWT_SECRET}
       - GMAPS_API_KEY=${GMAPS_API_KEY}
+      - API_REDIS_SERVICE_PORT=${API_REDIS_SERVICE_PORT}
+      - API_REDIS_SERVICE_HOST=${API_REDIS_SERVICE_HOST}
     depends_on:
       - postgresql
+      - redis
+
+  redis:
```
tiagojsag, line 45: please review this. This is twice as complex as it needs to be. copy-pasting from the internet should be done with care ;)

tiagojsag (changes requested):
Mostly reviewed the devops/gh actions/project config part. Also: I saw a lot of relative imports on the tests, especially for src code. those are evil AF, please refactor

docker-compose.yml
alexeh: Ah yes, ideas of a mad man when feeling alone in the darkness. I totally forgot this, thanks

api/test/jest-config.json
alexeh: It is meant to help with memory usage by compiling code on a module basis instead of this is sparta style.
I found it helpful on my machine so letting this here.

Let me know if you want to remove it anyway

api/package.json
alexeh: I'd say:
Increasing heap size for node: yes
expose garbage collector: no
call jest from node modules (to avoid some global version running): yes
tracking memory usage: yes (I found it very useful and a secret tools tha
the rest of the flags are helping with mem usage in testing and clearing up some async jobs so I'd say: yes

docker-compose.yml
alexeh: This is the best I could simplify. Let me know how this can be improved please :)

api/test/jest-config.json
tiagojsag: it seems to have some drawbacks: https://huafu.github.io/ts-jest/user/config/isolatedModules
have you looked into those?

api/src/modules/import-data/workers/import-data.producer.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface ExcelImportJob {
+  xlsxFilePath: string;
+  userId: string;
+}
+
+export class ImportQueueName extends TinyTypeOf<string>() {}
+
+export const importQueueName: ImportQueueName = new ImportQueueName(
+  'excel-import',
+);
```
alexeh, line 15: Not a very smart use of the recently added tool, just a try. This could be just a common uppercase constant

api/test/jest-config.json
alexeh: Yes I did, project tsc keeps working on the IDE so we still have type safety writing tests. I fail recognising more caveats here so let me know if something smells bad so I can change it back

api/src/modules/import-data/import-data.controller.ts
```diff
@@ -35,10 +32,8 @@ export class ImportDataController {
   async importSourcingRecords(
     @UploadedFile() xlsxFile: Express.Multer.File,
   ): Promise<any> {
-    try {
-      return await this.importDataService.importSourcingRecords(xlsxFile.path);
-    } catch (err) {
-      throw new BadRequestException(err);
-    }
+    // TODO: Add userId once auth is in place
+    const userId: string = '2a833cc7-5a6f-492d-9a60-0d6d056923ea';
```
alexeh, line 36: This could be a good usage for a tiny type, but messes with some DTOs, entity itself and potentially a middleware we will have after guarding some controller/endpoints

Let's discuss further

api/test/integration/import-data/xlsx-uploads/sourcing-data-import.spec.ts
```diff
@@ -58,7 +58,7 @@ jest.mock('config', () => {
   return config;
 });
 
-describe('Sourcing Data import', () => {
+describe.skip('Sourcing Data import', () => {
```
alexeh, line 61: Skipped in this PR. Sourcing Data Import tests updated in: https://github.com/Vizzuality/landgriffon/pull/204

api/src/modules/import-data/workers/import-data.producer.ts
tiagojsag: you mean the queue name? use a Symbol

api/test/e2e/h3-data/h3-impact-map.spec.ts
```diff
@@ -41,7 +41,7 @@ describe('H3 Data Module (e2e) - Impact map', () => {
   afterAll(async () => {
     await dropH3DataMock([fakeTable]);
     await deleteImpactMapMockData();
-    return app.close();
+    await app.close();
```
tiagojsag, line 44: these are functionally equivalent AFAIK. could you explain the change please?

api/test/integration/import-data/xlsx-uploads/import-data-produce-jobs.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+  test('When loadXlsxFile is called with required file data and a userId, but the Job can not be added to the queue, a error should be displayed', async () => {
+    await bootstrapTestingApp(importQueueFail);
+    const fileData = {
+      path: 'filepath',
+      filename: 'filename',
+    } as Express.Multer.File;
+    const userId = '29b47625-9ffa-45a9-bfce-7a75abeb820e';
+    expect.assertions(1);
+    try {
```
tiagojsag, line 52: did you try something like https://stackoverflow.com/questions/46042613/how-to-test-the-type-of-a-thrown-exception-in-jest instead? seems cleaner and more intentional

api/test/integration/import-data/xlsx-uploads/sourcing-data-import.spec.ts
tiagojsag: Add a comment with a @TODO explaining why we are skipping it, so we know when we can restore it

api/test/integration/import-data/xlsx-uploads/import-data-produce-jobs.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    await moduleFixture.close();
+  });
+
+  test('When loadXlsxFile is called with required file data and a userId, Then a job should be pushed to the queue', async () => {
+    await bootstrapTestingApp(importQueue);
+    const fileData = { path: 'filepath' } as Express.Multer.File;
+    const userId = '29b47625-9ffa-45a9-bfce-7a75abeb820e';
+    await importDataService.loadXlsxFile(userId, fileData);
+
+    expect(queue.add).toHaveBeenCalledWith('excel-import-job', {
```
tiagojsag, line 39: This test leaves data in the queue. You have two options:
- either cleanup all queues on afterEach()
- Do something like https://github.com/Vizzuality/marxan-cloud/commit/0a87118dd45216ea6db3e8b5eee39fbb9a762057

tiagojsag (changes requested):

api/test/e2e/h3-data/h3-impact-map.spec.ts
alexeh: Because we are awaiting to app.close() in 90% of the tests suites, I think it's cleaner to be consistent.

api/test/integration/import-data/xlsx-uploads/import-data-produce-jobs.spec.ts
alexeh: Event tho we are getting the queue by its token name after starting the module, what we are injecting as the queue instance before starting the test module is a mock object which is a just a jest spy, therefore this test does not add data to the queue. just triple checked locally again

What I had in mind and thought it would be a great task for @yulia-bel is setting up a test redis-db as we have with for the persistence layer so she can play a bit with redis. One of the reasons also why I created a different tasks for testing the events to let her play and get used to the queue events

api/test/integration/import-data/xlsx-uploads/import-data-produce-jobs.spec.ts
alexeh: I honestly prefer my current approach as it lets you do more stuff with exceptions without relying on the jest API
Also I just gave it a try and for some reasons the expect block does not catch the exception, tried with .toThrow, toTrowError and a couple of methods more.

So taking in account that the test is accurate and for the sake of pragmatism, I'd suggest to leave it as is, because the flow will most likely change. We can add this to lightweight tech debt if you are OK with that

api/test/e2e/h3-data/h3-impact-map.spec.ts
tiagojsag: we should probably prefer the return version, but that's outside of scope here. leave it as you have it :)

api/test/integration/import-data/xlsx-uploads/import-data-produce-jobs.spec.ts
tiagojsag: I missed the mock :facepalm: 

Not sure I understand where you want to set up a redis for testing. AFAIK we already have it on the servers, in CI and in the docker-compose for local execution, so not sure where it's missing. Let's chat about it, maybe

api/src/modules/import-data/workers/import-data.producer.ts
alexeh: I gave it a try, and it does not work, (even after converting to value as a string)

1.  Even tho I saw the marxan approach, the decorators and the BullModule we are using need that value as string
2. After converting the value to a string resulting 'Symbol(queueName)' there is a exception thrown when API starts.
3. Taking in account the constant statement and the primitive nature of the string, this should be inmutable as well

Thoughts?

api/test/integration/import-data/xlsx-uploads/import-data-produce-jobs.spec.ts
alexeh: Just a thought, even tho I have experience with queue stuff like activeMQ and SQS, I am new to redis. I saw it provides a way to use different 'databases' in the **same instance,** and also NestJS provides a way to pass different config values... so I thought we can already do it and that could be useful for some reason. I haven't think it deeply to be hones

tiagojsag: let's talk about this on Monday

PR #204, opened 2022-01-05
chore/LANGRIF-391 - Track Import Job Events

alexeh:
This PR adds:

Second piece of https://vizzuality.atlassian.net/browse/LANDGRIF-236

**Please NOTE**
This is a fist version of tracking async jobs related to the spreadsheet import.
There is a need of talking about how we actually want to handle it: which data store, which errors store (we might want to let the user know some of them, not all)...
And overall a lot of stuff to decide

This PR keeps tracks of the jobs by adding a new module: job-events. It is not meant to add an event-driven layer. (at least yet)

Updating job events relies on the events that the underlying queue emits and we can listen to. This could be done in a lot of different ways.

Testing the sourcing-import data now bypasses the queue system itself, we test that we can push a job to the queue (done in the first piece of this task: https://github.com/Vizzuality/landgriffon/pull/202)

And here the test subject is the sourcing-import service itself.

Implementation of an endpoint to retrieve job statuses and related tests is left to a different task as there will be changes about how we want to handle the data, what we want to expose to the user and etc


yulia-bel (approved):

api/test/integration/import-data/xlsx-uploads/import-data-produce-jobs.spec.ts
```diff
@@ -5,13 +5,18 @@ import { ImportDataModule } from 'modules/import-data/import-data.module';
 import { getQueueToken } from '@nestjs/bull';
 import { importQueueName } from 'modules/import-data/workers/import-data.producer';
 import { ImportDataService } from 'modules/import-data/import-data.service';
+import { TasksRepository } from '../../../../src/modules/tasks/tasks.repository';
```
tiagojsag, line 8: image

at least when importing from src to test

tiagojsag:
Please rebase to address the conflict and look into the failing tests

PR #206, opened 2022-01-10
Add authentication

asuramaru1:
Add authentication to the API.
JIRA :https://vizzuality.atlassian.net/browse/LANDGRIF-410

### DIFF
--- a/api/config/custom-environment-variables.json
+++ b/api/config/custom-environment-variables.json
@@ -29,5 +29,11 @@
     "sizeLimit": "FILE_SIZE_LIMIT",
     "storagePath": "FILE_STORAGE_PATH"
   },
-  "geolocation": {"gmapsApiKey": "GMAPS_API_KEY" }
+  "geolocation": {"gmapsApiKey": "GMAPS_API_KEY" },
+  "password": {
+    "minLength": "PASSWORD_MIN_LENGTH",
+    "includeNumerics": "PASSWORD_INCLUDE_NUMERICS",
+    "includeUpperCase": "PASSWORD_INCLUDE_UPPER_CASE",
+    "includeSpecialCharacters": "PASSWORD_INCLUDE_SPECIAL_CHARACTERS"
+  }
 }
--- a/api/config/custom-environment-variables.json
+++ b/api/config/custom-environment-variables.json
@@ -3,10 +3,18 @@
     "port": "SERVER_PORT"
   },
   "auth": {
+    "requireUserAuth": "REQUIRE_USER_AUTH",
     "requireUserAccountActivation": "REQUIRE_USER_ACCOUNT_ACTIVATION",
+    "signUpIsPublic": "SIGN_UP_IS_PUBLIC",
     "jwt": {
       "expiresIn": "JWT_EXPIRES_IN",
       "secret": "JWT_SECRET"
+    },
+    "password": {
+      "minLength": "PASSWORD_MIN_LENGTH",
+      "includeNumerics": "PASSWORD_INCLUDE_NUMERICS",
+      "includeUpperCase": "PASSWORD_INCLUDE_UPPER_CASE",
+      "includeSpecialCharacters": "PASSWORD_INCLUDE_SPECIAL_CHARACTERS"
     }
   },
   "import": {
@@ -29,11 +37,5 @@
     "sizeLimit": "FILE_SIZE_LIMIT",
     "storagePath": "FILE_STORAGE_PATH"
   },
-  "geolocation": {"gmapsApiKey": "GMAPS_API_KEY" },
-  "password": {
-    "minLength": "PASSWORD_MIN_LENGTH",
-    "includeNumerics": "PASSWORD_INCLUDE_NUMERICS",
-    "includeUpperCase": "PASSWORD_INCLUDE_UPPER_CASE",
-    "includeSpecialCharacters": "PASSWORD_INCLUDE_SPECIAL_CHARACTERS"
-  }
+  "geolocation": {"gmapsApiKey": "GMAPS_API_KEY" }
 }
--- a/api/config/default.json
+++ b/api/config/default.json
@@ -7,10 +7,18 @@
     ]
   },
   "auth": {
+    "requireUserAuth": true,
     "requireUserAccountActivation": true,
+    "signUpIsPublic": true,
     "jwt": {
       "expiresIn": "2h",
       "secret": null
+    },
+    "password": {
+      "minLength": 6,
+      "includeNumerics": false,
+      "includeUpperCase": false,
+      "includeSpecialCharacters": false
     }
   },
   "import": {
@@ -38,11 +46,5 @@
   },
   "geolocation": {
     "gmapsApiKey": null
-  },
-  "password": {
-    "minLength": 6,
-    "includeNumerics": false,
-    "includeUpperCase": false,
-    "includeSpecialCharacters": false
   }
 }
--- a/api/config/default.json
+++ b/api/config/default.json
@@ -38,5 +38,11 @@
   },
   "geolocation": {
     "gmapsApiKey": null
+  },
+  "password": {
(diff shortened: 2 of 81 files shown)


api/test/e2e/admin-regions/admin-regions-get-trees.spec.ts
```diff
@@ -36,6 +38,16 @@ describe('AdminRegions - Get trees', () => {
     );
     await app.init();
 
+    await request(app.getHttpServer())
```
tiagojsag, line 41: Extract this into a separate function, so you don't have to repeat the same block of code in every file like you did.
Also: instead of using the sign-up endpoint, manually inject a user into the database, as it's easier to extend

api/config/default.json
```diff
@@ -38,5 +38,11 @@
   },
   "geolocation": {
     "gmapsApiKey": null
+  },
+  "password": {
```
tiagojsag, line 42: add these to the custom-environment-variables.json file too, so they can more easily be configured

api/src/decorators/password-validator.decorator.ts
```diff
@@ -0,0 +1,30 @@
...
+  ValidatorConstraintInterface,
+  ValidationArguments,
+} from 'class-validator';
+import * as config from 'config';
+
+@ValidatorConstraint({ name: 'PasswordValidation', async: false })
+export class PasswordValidation implements ValidatorConstraintInterface {
+  validate(password: string, args: ValidationArguments) {
+    const minLength: number = config.get('password.minLength');
+    const includeUpperCase: boolean = config.get('password.includeUpperCase');
```
tiagojsag, line 12: for all values being loaded from config, handle the case when they may be a string. Use something like:

```
`${config.get('includeUpperCase')}`.toLowerCase() !== 'false'
```

tiagojsag (changes requested):
Not sure if we wrote this down anywhere (apologies if we didn't) but the spec is that all but a couple of endpoint should require auth. So it's easier and cleaner to deploy the auth guard as a global guard, and exclude the few endpoints. you can find an example of that here: https://dev.kuffel.io/nestjs-bypass-endpoint-auth-guards/

api/src/decorators/password-validator.decorator.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    return !(includeSpecialCharacters && !/(?=.*[!@#$%^&*])/.test(password));
+  }
+
+  defaultMessage(args: ValidationArguments): string {
+    const password: string = args.value;
+    let message: string = '';
+
+    if (!PasswordValidation.minLengthCheck(password))
+      message += `Password too short , minimal length is ${config.get(
```
tiagojsag, line 49:
```suggestion
      message += `Password too short, minimum length is ${config.get(
```

api/src/decorators/password-validator.decorator.ts
```diff
@@ -0,0 +1,67 @@
...
+    const includeSpecialCharacters: boolean =
+      `${config.get('password.includeSpecialCharacters')}`.toLowerCase() ===
+      'true';
+
+    return !(includeSpecialCharacters && !/(?=.*[!@#$%^&*])/.test(password));
+  }
+
+  defaultMessage(args: ValidationArguments): string {
+    const password: string = args.value;
+    let message: string = '';
```
tiagojsag, line 46: make message an array of string, append elements to it inside each `if`, and join() in the end

tiagojsag (changes requested):
left a couple or notes. Also, going forward, please avoid meaningless commit messages like "refactor". Anyone who looks at that message will not understand what changes you did, or why you did them

api/config/custom-environment-variables.json
```diff
@@ -29,5 +29,11 @@
...
     "storagePath": "FILE_STORAGE_PATH"
   },
-  "geolocation": {"gmapsApiKey": "GMAPS_API_KEY" }
+  "geolocation": {"gmapsApiKey": "GMAPS_API_KEY" },
+  "password": {
+    "minLength": "PASSWORD_MIN_LENGTH",
+    "includeNumerics": "PASSWORD_INCLUDE_NUMERICS",
+    "includeUpperCase": "PASSWORD_INCLUDE_UPPER_CASE",
+    "includeSpecialCharacters": "PASSWORD_INCLUDE_SPECIAL_CHARACTERS"
+  }
```
alexeh, line 38: I think this env vars should be added also to the docker-compose setup for team members running this on docker.

Also it would be great adding them to the dotenv skeleton (.env.default) so people can notice there are new env vars

tiagojsag: we have reasonable defaults already, so no need to add these to the docker-compose or .env skeleton. But I do think we should have a list documenting all supported config env vars. Maybe add a section to the README documenting the vars you just added.

api/src/modules/users/users.controller.ts
```diff
@@ -68,6 +68,18 @@ export class UsersController {
     return this.service.serialize(results.data, results.metadata);
   }
 
+  @ApiOperation({ description: 'Create new user' })
+  @ApiCreatedResponse({ description: 'User created successfully' })
+  @ApiOkResponse({
+    type: User,
+  })
+  @Post()
```
alexeh, line 76: Should this route be more defined, like: /users/create ?

tiagojsag: what would be the upside of that vs current implementation?

alexeh: I feel like a endpoint defined us 
`user/create`
is more meaningful than
'user'

asuramaru1: In most REST API practices , POST , GET , DELETE ...  define what endpoint does to a resource .POST is creating new resource  so I think adding extra /create is redundant here  , same way as DELETE /{resource}/id  is used instead of DELETE /{resource}/id/delete . @alexeh please let me know what you think about this

alexeh: My comment was mostly because of the spec of adding a create user endpoint, but after your reasoning I changed my mind.
Thank you for the detailed explanation

api/src/modules/users/users.service.ts
```diff
@@ -175,4 +175,8 @@ export class UsersService extends AppBaseService<
       );
     }
   }
+
+  async createUser(createUserDTO: CreateUserDTO): Promise<Partial<User>> {
+    return await this.authenticationService.createUser(createUserDTO);
```
tiagojsag, line 180:
```suggestion
    return this.authenticationService.createUser(createUserDTO);
```

alexeh (approved):
Thank you @asuramaru1 !

yulia-bel (approved):

api/src/modules/users/users.controller.ts
tiagojsag: +1 @asuramaru1

tiagojsag (changes requested):
I think the implementation in itself if fine as it is. I think that we are missing the tests to cover the functionality that was added - @asuramaru1 can you look into that, please? There should be an example in the test already of how you can use fake config values to tweak the password verification variables, and add tests for those.

Another thing that is NOT part of this task, but that we need to add before merging this, is a way to add users. Right now, you need to be logged in to create a user, and since there are no users by default.... we'll effectively lock the platform if we merge this. So let's keep this open for now, even once we have the tests in place.

asuramaru1:
@tiagojsag thanks! I'll add tests to cover password validation for all cases .
Also , sign up works  as well , could you please explain why this will lock the platform ?
And thanks for thorough review

alexeh:
SignUp and SignIn are public routes so we can create users through the API ( I also tested it).
Also we figured out a way to not block the api until the client can handle auth

tiagojsag (approved):

tiagojsag:
@alexeh please coordinate when best to merge this, so we dont lock people out of the platform

PR #209, opened 2022-01-11
Fixed materials and suppliers for more filters selections

davidsingal:


alexeh (approved):
Just had a look on the deployed version, looks working smoothly now, thank you!

PR #210, opened 2022-01-12
Update years fallback strategy for impact data

yulia-bel:
This PR implements the updates of recent changes to years fallback strategy for Impact maps:
- In Indicator Record service in case of 'fallback' the year of material H3 is now found in accordance with latest fallback strategy (before it was getting the latest available year)
- In the maps responses the format of the information about material H3 data years has changed - since Impact map can be requested for multiple materials, now, in metadata, materials years are given as an array of objects, each object containing material name, material type (harvest or producer) and material year.
- methods helping to retrieve years moved from H3 Data Service to H3FilterYearsByLayerService (as Alex suggested before)
- Tests updates

### DIFF
--- a/api/src/modules/h3-data/h3-data.service.ts
+++ b/api/src/modules/h3-data/h3-data.service.ts
@@ -10,7 +10,10 @@ import { H3Data, H3IndexValueData } from 'modules/h3-data/h3-data.entity';
 import { MaterialsService } from 'modules/materials/materials.service';
 import { IndicatorsService } from 'modules/indicators/indicators.service';
 import { UnitConversionsService } from 'modules/unit-conversions/unit-conversions.service';
-import { H3MapResponse } from 'modules/h3-data/dto/h3-map-response.dto';
+import {
+  H3MapResponse,
+  MaterialsH3DataYears,
+} from 'modules/h3-data/dto/h3-map-response.dto';
 import {
   Indicator,
   INDICATOR_TYPES,
@@ -20,6 +23,7 @@ import { H3FilterYearsByLayerService } from 'modules/h3-data/services/h3-filter-
 import { GetImpactMapDto } from 'modules/h3-data/dto/get-impact-map.dto';
 import { MaterialsToH3sService } from 'modules/materials/materials-to-h3s.service';
 import { MATERIAL_TO_H3_TYPE } from 'modules/materials/material-to-h3.entity';
+import { Material } from 'modules/materials/material.entity';
 
 /**
  * @debt: Check if we actually need extending nestjs-base-service over this module.
@@ -164,18 +168,11 @@ export class H3DataService {
      * @note To generate a Risk Map, harvest data and h3Data by indicatorId are required
      */
 
-    const availableIndicatorYears: number[] =
-      await this.getAvailableYearsForH3IndicatorData(indicatorId);
-
-    let indicatorDataYear: number | undefined =
-      availableIndicatorYears.includes(year)
-        ? year
-        : availableIndicatorYears.find((el: number) => el < year);
-
-    if (!indicatorDataYear)
-      indicatorDataYear = availableIndicatorYears
-        .reverse()
-        .find((el: number) => el > year);
+    const indicatorDataYear: number | undefined =
+      await this.filterYearsByLayerService.getH3IndicatorYearForCalculations(
+        indicatorId,
+        year,
+      );
 
     if (!indicatorDataYear)
       throw new NotFoundException(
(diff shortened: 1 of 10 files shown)


tiagojsag:
tests be failing

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -410,30 +408,46 @@ export class H3DataService {
       getImpactMapDto.supplierIds,
     );
 
+    const materialsH3DataYears: MaterialsH3DataYears[] = [];
+
+    if (getImpactMapDto.materialIds?.length)
+      for (const materialId of getImpactMapDto.materialIds) {
+        const material: Material = await this.materialService.getMaterialById(
```
alexeh, line 415: This makes roundtrips to the DB as many times as materials are requested from the consumer being a hit to performance. Given the data-intensive nature of this product, I think we should try to avoid as many DB calls as possible. Do you think we can improve this?

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -410,30 +408,46 @@ export class H3DataService {
...
+          {
+            materialName: material.name,
+            materialDataYear: materialHarvestH3DataYear,
+            materialDataType: MATERIAL_TO_H3_TYPE.HARVEST,
+          },
+          {
+            materialName: material.name,
+            materialDataYear: materialProducerH3DataYear,
+            materialDataType: MATERIAL_TO_H3_TYPE.PRODUCER,
+          },
```
alexeh, line 440: Are our beloved frontends aware that you are retrieving this info when sending a map back? I understand the reasoning of adding this info, but I'd like to know if this is going to be actually used, as any kind of map is already a huge payload and I think we should try to retrieve just the exact needed amount of data

api/test/e2e/h3-data/h3-impact-map.spec.ts
```diff
@@ -267,6 +288,8 @@ describe('H3 Data Module (e2e) - Impact map', () => {
       expect(response.body.metadata).toEqual({
         quantiles: [617, 617, 617, 617, 617, 617, 617],
         unit: 'tonnes',
+        indicatorDataYear: 2020,
+        materialsH3DataYears: [],
```
alexeh, line 292: On top of: https://github.com/Vizzuality/landgriffon/pull/210/files#r784572397

I guess this is the use case for when we actually have matching years for materials. 
Should we add an empty array in this case?

api/src/modules/h3-data/h3-data.service.ts
```diff
@@ -191,27 +188,25 @@ export class H3DataService {
...
-      harvestDataYear = availableHarvestDataYears
-        .reverse()
-        .find((el: number) => el > year);
-
     if (!harvestDataYear)
       throw new NotFoundException(
         `There is no H3 Harvest data registered for Material with ID ${materialId} for year ${year} or any other year`,
       );
 
+    materialsH3DataYears.push();
```
alexeh, line 209: What is this for?

api/src/modules/h3-data/services/h3-filter-years-by-layer.service.ts
```diff
@@ -89,4 +92,62 @@ export class H3FilterYearsByLayerService {
...
+
+    if (!materialDataYear)
+      materialDataYear = availableH3DataYears
+        .reverse()
+        .find((el: number) => el > year);
+
+    return materialDataYear;
+  }
+
+  async getH3IndicatorYearForCalculations(
```
alexeh, line 117: I have some suggestions here:

1. At a very personal level, I find the name a bit weird to understand what is it's purpose, as a person who has not been involved lately in this fallback strategy. Either a jsdocs documentation or maybe a more meaningful name (or both?) as `getClosestAvailableYearForMaterialH3` would be better. Thoughts?

2. In line 103 you are using  a method which is basically a wrapper for a method implemented in h3-data-repository, with no additional logic in it. Since this service receives the singleton instance of said repository, I think you can directly access said method avoiding adding 
 -  getAvailableYearsForH3MaterialData
 - getAvailableYearsForH3IndicatorData
 methods.
 
 Also, but this is out of this PR's scope, I think it would be a good idea to handle the error for when no year could be found inside these methods, which is currently happening outside.

alexeh (changes requested):
Thanks @yulia-bel !

I added some (I hope) improvements suggestions and a few questions I'd like to clarify.

Also, since you used the service related to years to do stuff with years (which is great IMO) you also extended it's functionallity (which is x2 great) but now I found the naming a bit inaccurate

So when it was first born, this service could only do `filter-years-by-layer` but now it does more, so I feel like a renaming to something 'more general' would be great

Thanks!

api/src/modules/h3-data/h3-data.service.ts
yulia-bel: My idea was to include in the response all the relevant information and then present it to David and Elena for feedback. In terms of information, in my opinion, it is correct if we want to have full transparency on the whole years situation. Impact map can be requested for multiple materials, lets say user wants impact map for cocoa and cereal for year 2020, but the calculations were based on the closest available years, which in theory can be different for each material and type (crop o producer). Let's say cocoa impact calculations are based on 2010 producer h3data, 2009 crop h3data, and cocoa impact calculations are based on 2005 producer h3data and 2007 crop h3data. So we need to inform user about all those years for each requested material and type. But again final decision shall be made by FE / Science

api/src/modules/h3-data/services/h3-filter-years-by-layer.service.ts
yulia-bel: Thanks a lot for suggestions - all implemented.

api/src/modules/h3-data/services/h3-data-years.service.ts
```diff
@@ -89,4 +92,53 @@ export class H3FilterYearsByLayerService {
...
+
+    if (!materialDataYear)
+      materialDataYear = availableH3DataYears
+        .reverse()
+        .find((el: number) => el > year);
+
+    return materialDataYear;
+  }
+
+  async getClosestAvailableYearIndicatorH3(
```
alexeh, line 124:
```suggestion
  async getClosestAvailableYearForIndicatorH3(
```

alexeh (approved):
Thank you @yulia-bel nice one! 

I just added a minor 'typo' correction suggestion. Also, as you changed the year service file name, I think it would be great to rename the class itself.

But feel free to go forward with this from my side. I can rename it later myself

Thank you nice job!

PR #211, opened 2022-01-14
Fix: LANDGRIF-436 - Handle Redis connection at init

alexeh:
This PR adds:

A Queue event listener which throws an unhandled error if the API cannot connect to a queue in Redis. Will also crash if the queue goes down during API runtime

TF k8s setup adding redis host env var to api deployment

**NOTE:**
OnGlobalQueueError won't do the job as expected. For now I think we can live with this until we implement more queues on the system. If more queues are added, most likely we will need a more generalistic approach to handle the connection, like a global consumer.

Works nice and no more deps are added to the project


kubernetes/modules/secrets/main.tf
```diff
@@ -81,3 +81,16 @@ resource "kubernetes_secret" "db_secret" {
     DB_SYNCHRONIZE = "true"
   }
 }
+
+resource "kubernetes_secret" "redis_secret" {
```
tiagojsag, line 85: no need to create a new secret just for this, reuse the existing db_secret

tiagojsag (changes requested):

tiagojsag (approved):

PR #212, opened 2022-01-18
add prototype for getting distributed impact map

elpamart:
This PR contains a wip prototype in a jupyter notebook for retrieving all the sourcing records and calculate the total water impact in a distributed way. The query needs to be optimized.


aagm (approved):
We have discussed past week about this, in terms of methodology i will ping @ikerey, Looking at the solution part of the queries, we might need to change a bit the db model in this case due to performance

PR #213, opened 2022-01-18
Add Tasks endpoint

yulia-bel:
This PR adds:
- Basic full CRUD for Tasks module.
- Creates new dto for updating a Task via controller, since the existing one has limited properties and is already used in the import process, probably we can unite them at some point.
- The properties which user can update in a Task or add, when creating new Tasks, are quite random, it can be easily changed if suggested
- Test for Tasks endpoints
Filtering tasks by User Id is not implemented yet, waiting for the Auth PR to be merged

### DIFF
--- a/api/src/modules/tasks/dto/update-task-with-controller.dto.ts
+++ b/api/src/modules/tasks/dto/update-task-with-controller.dto.ts
@@ -1,0 +1,23 @@
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { IsNotEmpty, IsOptional, IsString, IsUUID } from 'class-validator';
+import { TASK_STATUS } from 'modules/tasks/task.entity';
+
+export class UpdateTaskWithControllerDto {
+  @ApiProperty()
+  @IsNotEmpty()
+  @IsUUID()
+  taskId: string;
+
+  @ApiProperty()
+  @IsNotEmpty()
+  @IsString()
+  newStatus: TASK_STATUS;
+
+  @ApiPropertyOptional()
+  @IsOptional()
+  newData?: Record<string, any>;
+
+  @ApiPropertyOptional()
+  @IsOptional()
+  newErrors?: Error;
+}
--- a/api/src/modules/tasks/tasks.controller.ts
+++ b/api/src/modules/tasks/tasks.controller.ts
@@ -1,4 +1,104 @@
-import { Controller } from '@nestjs/common';
+import {
+  Body,
+  Controller,
+  Delete,
+  Get,
+  Param,
+  Post,
+  Put,
+  UsePipes,
+  ValidationPipe,
+} from '@nestjs/common';
+import {
+  ApiBadRequestResponse,
+  ApiForbiddenResponse,
+  ApiNotFoundResponse,
+  ApiOkResponse,
+  ApiOperation,
+  ApiTags,
+  ApiUnauthorizedResponse,
+} from '@nestjs/swagger';
+import {
+  FetchSpecification,
+  ProcessFetchSpecification,
+} from 'nestjs-base-service';
+import { PaginationMeta } from 'utils/app-base.service';
+import { Task, taskResource } from 'modules/tasks/task.entity';
+import { TasksService } from 'modules/tasks/tasks.service';
+import { UpdateTaskWithControllerDto } from 'modules/tasks/dto/update-task-with-controller.dto';
+import { CreateTaskDto } from 'modules/tasks/dto/create-task.dto';
 
-@Controller('tasks')
-export class TasksController {}
+@Controller('/api/v1/tasks')
+@ApiTags(taskResource.className)
+export class TasksController {
+  constructor(protected readonly taskService: TasksService) {}
+
+  @ApiOperation({ description: 'Find all tasks' })
+  @ApiOkResponse({ type: Task })
+  @ApiBadRequestResponse()
+  @ApiUnauthorizedResponse()
+  @ApiForbiddenResponse()
+  @Get()
+  async findAll(
+    // To implement once Auth is merged:
+    //@Req() request: Request,
+    @ProcessFetchSpecification({
+      allowedFilters: taskResource.columnsAllowedAsFilter,
+    })
+    fetchSpecification: FetchSpecification,
+  ): Promise<Task> {
+    // To implement once Auth is merged:
+    // fetchSpecification.filter
+    //   ? (fetchSpecification.filter.createdBy = [request.user.id])
+    //   : (fetchSpecification.filter = {
+    //       createdBy: [request.user.id],
+    //     });
+    const results: {
+      data: (Partial<Task> | undefined)[];
+      metadata: PaginationMeta | undefined;
+    } = await this.taskService.findAllPaginated(fetchSpecification);
+    return this.taskService.serialize(results.data, results.metadata);
+  }
+
+  @ApiOperation({ description: 'Find task by id' })
+  @ApiOkResponse({ type: Task })
+  @ApiBadRequestResponse()
+  @ApiUnauthorizedResponse()
+  @ApiForbiddenResponse()
+  @Get(':id')
+  async findOne(@Param('id') id: string): Promise<Task> {
+    return await this.taskService.serialize(await this.taskService.getById(id));
+  }
+
+  @ApiOperation({ description: 'Create a Task' })
+  @ApiOkResponse({ type: Task })
+  @ApiBadRequestResponse({
+    description: 'Bad Request. Incorrect or missing parameters',
+  })
+  @Post()
+  @UsePipes(ValidationPipe)
+  async create(@Body() dto: CreateTaskDto): Promise<Task> {
+    return await this.taskService.serialize(await this.taskService.create(dto));
+  }
+
+  @ApiOperation({ description: 'Updates a task' })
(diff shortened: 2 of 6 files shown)


api/src/modules/tasks/dto/update-task-with-controller.dto.ts
```diff
@@ -0,0 +1,23 @@
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { IsNotEmpty, IsOptional, IsString, IsUUID } from 'class-validator';
+import { TASK_STATUS } from 'modules/tasks/task.entity';
+
+export class UpdateTaskWithControllerDto {
+  @ApiProperty()
+  @IsNotEmpty()
+  @IsUUID()
+  taskId: string;
```
tiagojsag, line 9:
```suggestion
  taskId!: string;
```

api/src/modules/tasks/dto/update-task-with-controller.dto.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  @ApiProperty()
+  @IsNotEmpty()
+  @IsString()
+  newStatus: TASK_STATUS;
```
tiagojsag, line 14:
```suggestion
  newStatus?: TASK_STATUS;
```

tiagojsag: overall I think you need to account for updates on things other than status (even if those are a bit strange)

api/src/modules/tasks/dto/update-task-with-controller.dto.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  @ApiPropertyOptional()
+  @IsOptional()
+  newData?: Record<string, any>;
+
+  @ApiPropertyOptional()
+  @IsOptional()
+  newErrors?: Error;
```
tiagojsag, line 22: I suspect the type here is wrong

api/src/modules/tasks/tasks.controller.ts
```diff
@@ -1,4 +1,104 @@
...
+  @Get()
+  async findAll(
+    // To implement once Auth is merged:
+    //@Req() request: Request,
+    @ProcessFetchSpecification({
+      allowedFilters: taskResource.columnsAllowedAsFilter,
+    })
+    fetchSpecification: FetchSpecification,
+  ): Promise<Task> {
+    // To implement once Auth is merged:
```
tiagojsag, line 50: not really sure what you are trying to do here.

api/test/e2e/tasks/tasks.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+      const task1: Task = new Task();
+      task1.createdBy = '2a833cc7-5a6f-492d-9a60-0d6d056923ea';
+      task1.type = TASK_TYPE.SOURCING_DATA_IMPORT;
+      task1.status = TASK_STATUS.PROCESSING;
+      task1.data = {
+        filename: 'fakeFile.xlsx',
+      };
+      await task1.save();
+
+      const task2: Task = new Task();
```
tiagojsag, line 178: see the `entity-mock.ts` file and use something like that.

tiagojsag (changes requested):
I suspect your DTOs have a few things off around required/optional attributes, especially when compared to the actual entity def. @alexeh may know more

api/src/modules/tasks/dto/update-task-with-controller.dto.ts
alexeh: Actually the method taking care of updating a task's status if any error is thrown, consumes an error to then push an object with it's name as key and the message as value.

But this happens within the API, so I think that typing a new error as Error when it is meant to be updated via controller makes sense.

As a note I'd like to add that I think we should not allow to update an error on a task via controller as any possible error would happen withing the API.
Thoughts on this?

api/src/modules/tasks/tasks.controller.ts
alexeh: I believe the intention here is to add the ability to return tasks created by a given user, with the ID coming in the request once auth is done.
As mentioned before, the proper approach to this would be based on a role, but I think we can work with this. 
Right now the API returns a task once a import job is pushed to the queue and we also provide a way to get that task by it's ID, but I think we should keep this as well

api/src/modules/tasks/dto/update-task-with-controller.dto.ts
alexeh: I agree. I believe here the newStatus is required since the method UPDATING a status **within the api** requires a new status, but if we want to let a user/admin update a task, we should allow to update another fields as well

yulia-bel: so now, when updating Task, taskId is required, rest is optional for user to update (change status, add data, add errors). This endpoint doesn't allow to add logs, since it is not part of the updateImportJobEvent service used in the endpoint, I can add it to service or @alexeh, what do you think, probably I should create separate method for updating Task via controller to to mix it with the updates coming from import process?

api/src/modules/tasks/tasks.controller.ts
yulia-bel: yes, I added a GetUser decorator which is supposed to get user object from the request (after Auth is done), then the idea is to add user id to the 'filters' property of fetchSpecification object, which is passed to findAllPaginated() method of nestjs-base-service, so that it returns all Tasks created By the authenticated User

api/src/modules/authentication/get-user.decorator.ts
```diff
@@ -0,0 +1,9 @@
+import { createParamDecorator, ExecutionContext } from '@nestjs/common';
+import { User } from 'modules/users/user.entity';
+
+export const GetUser = createParamDecorator(
+  (_data, context: ExecutionContext): User => {
+    const request = context.switchToHttp().getRequest();
+    return request.user;
+  },
+);
```
alexeh, line 9: Since this is a global decorator, I think it should be added in
`api/src/decorators`

api/src/modules/tasks/tasks.controller.ts
```diff
@@ -1,4 +1,104 @@
...
+  async create(@Body() dto: CreateTaskDto): Promise<Task> {
+    return await this.taskService.serialize(await this.taskService.create(dto));
+  }
+
+  @ApiOperation({ description: 'Updates a task' })
+  @ApiNotFoundResponse({ description: 'Task not found' })
+  @ApiOkResponse({ type: Task })
+  @Put()
+  async update(
+    @Body(new ValidationPipe())
```
alexeh, line 89: not sure if you need to pass a new instance of the validation pipe, I think the Body decorator can instantiate it for you, like you are doing it in line 79 with the pipe. However really minor stuff

PS: I am not gonna be the one fighting for the async stuff xD

api/test/entity-mocks.ts
```diff
@@ -280,6 +281,17 @@ async function createSourcingRecord(
...
+async function createTask(status: TASK_STATUS) {
+  const task: Task = new Task();
+  task.createdBy = '2a833cc7-5a6f-492d-9a60-0d6d056923ea';
+  task.type = TASK_TYPE.SOURCING_DATA_IMPORT;
+  task.status = status;
+  task.data = {
+    filename: 'fakeFile.xlsx',
+  };
+  return task.save();
+}
```
alexeh, line 293: Plese could you add the ability to add additional / overwritte default data? Have a look how it is done with the rest of the entities in the mock utility, it has proven to be super useful more than once :)

yulia-bel: thanks a lot, all fixed :)

api/src/modules/tasks/dto/update-task-with-controller.dto.ts
tiagojsag: I would say don't worry too much about this endpoint, as it's more of a maintenance tool than anything else. Based on my experience, the most useful part will be updating status - in case a job gets stuck importing, and we need to fix it to unblock newer imports.

api/src/modules/tasks/dto/update-task-with-controller.dto.ts
tiagojsag: I agree with that last thought (and pretty much all of the rest).

api/src/modules/tasks/tasks.controller.ts
tiagojsag: ok, but I don't agree with returning just the tasks for a given user. At least for our only Task type right now, it is a process that affects all users, so I don't really see a motivation to hide other user's tasks. Maybe that's something we will need to add in the future, but for now I don't think it makes sense.

api/src/modules/tasks/dto/update-task-with-controller.dto.ts
yulia-bel: okay, so left possibility to update status and data via controller. If at some point we will need to be able to remove an error or logs, probably there should be a different, a bit more complex endpoint, since those columns are arrays of objects now..

api/src/modules/tasks/tasks.controller.ts
yulia-bel: okay, removed

tiagojsag (approved):

alexeh (approved):

PR #218, opened 2022-01-25
Targets Module

yulia-bel:
This PR adds Targets Module and tests.


api/src/modules/targets/targets.controller.ts
```diff
@@ -0,0 +1,107 @@
+import {
```
tiagojsag, line 1: image

tiagojsag: test if it works like it does now without them. In some scenarios, things change.

api/src/modules/targets/targets.service.ts
```diff
@@ -0,0 +1,38 @@
...
+    public readonly targetsRepository: TargetsRepository,
+  ) {
+    super(
+      targetsRepository,
+      targetResource.name.singular,
+      targetResource.name.singular,
+    );
+  }
+
+  // To Do
```
tiagojsag, line 31: todo? what' missing?

api/test/integration/import-data/xlsx-uploads/.~lock.base-dataset.xlsx#
```diff
@@ -0,0 +1 @@
+,yulia,omen,20.01.2022 17:46,file:///home/yulia/snap/libreoffice/239/.config/libreoffice/4;
```
tiagojsag, line 1: image

tiagojsag (changes requested):

api/test/integration/import-data/xlsx-uploads/.~lock.base-dataset.xlsx#
yulia-bel: cleaned the mess, sorry 😬

alexeh (commented):
Sorry for being late to this CR.

Looks like most of the stuff has been properly reviewed / addressed

I just left a minor comment, let me know what you think about

Thanks!

alexeh (approved):
LGTM, thank you @yulia-bel!

PR #220, opened 2022-01-26
Feature: LANDGRIF-388 Return user data for impact features

alexeh:
This PR adds:

3 new endpoint to retrieve Materials, AdminRegions and Suppliers imported by the spreadsheet

There are some caveats on this implementation: 

- The ORM does not provide a rightJoin feature to retrieve needed mapped entities straight from sourcing locations, so I am getting these IDs and retrieve the data from the entity table.
- The ORM also does not provide a way to find 1 to N roots and children (or VS) by IDs, so I am retrieving the list with findTrees so I can take advantage of let the user decide the depth and also clean data with no H3 (in case of materials)

We need to also handle the case for the client when a user hasn't load any data, I think this should be a separate task.
The client could use the absence of any task to know there is no impact data, i.e


api/src/modules/sourcing-locations/sourcing-locations.service.ts
```diff
@@ -83,4 +86,62 @@ export class SourcingLocationsService extends AppBaseService<
 
     return await this.sourcingLocationRepository.save(sourcingLocation as any);
   }
+
+  /**
+   * @description Retrieve Ids of all required Entities and its Parent's present in Sourcing-Locations
```
tiagojsag, line 91:
```suggestion
   * @description Retrieve Ids of all required Entities and its Parents present in Sourcing-Locations
```

PR #233, opened 2022-02-01
Landgrif 481 filter by range of years for table and chart

mluena:
**DESCRIPTION**

This PR disables options of years selection on the impact page.

Expected behavior:
- Map - user should be able to select any year.
- Chart and Table:
       - When entering user should see the full range of dates.
       - User only will be able to filter data by a minimum of 5 years window. The rest of the options should appear as disabled
       
**JIRA**
https://vizzuality.atlassian.net/browse/LANDGRIF-481?atlOrigin=eyJpIjoiZDU4NGMyZDMyMDI4NGQ1Yjk0YThiYmI1MDQ4MmU3YTQiLCJwIjoiaiJ9

### DIFF
--- a/client/src/components/select/component.tsx
+++ b/client/src/components/select/component.tsx
@@ -134,13 +134,15 @@ const ScenariosComparison: React.FC<SelectProps> = (props: SelectProps) => {
                 {optionsResult.map((option) => (
                   <Listbox.Option
                     key={option.value}
-                    className={({ active, selected }) =>
+                    className={({ active, selected, disabled }) =>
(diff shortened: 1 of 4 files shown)


davidsingal:
Could you avoid the 400 error when startYear and endYear is not selected?? 
image

My suggestion is to use `enable` param in the `useQuery`: https://react-query.tanstack.com/reference/useQuery

davidsingal (commented):
It looks good to me, I've only added a little request of change you can in the comments

PR #234, opened 2022-02-01
LANDGRIFF-244 Excel validation notebook

AngelArcones:
Validation of the input excel by field and by rows


data/notebooks/Lab/Excel_validation_logic.ipynb
```diff
@@ -0,0 +1,827 @@
...
+   "execution_count": 97,
+   "id": "14e73d61",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def location_validation(df):\n",
+    "    for l in range(len(df)):\n",
+    "        if 'country' in df.iloc[l]['location_type'].lower():\n",
+    "            if not pd.isna(df.iloc[l]['address']) or not pd.isna(df.iloc[l]['latitude']) or not pd.isna(df.iloc[l]['longitude']):\n",
+    "                print(f'LOCATION ERROR ON ENTRY {l+1}: address/latitude/longitude not allowed')\n",
```
elpamart, line 556: I wouldn't say we should provide this as an error but as a warning - the user can provide that information but it's not required/ used for the geocoding or maybe there is a case were the user can provide this information and maybe we can suggest the option of changing the location type?

data/notebooks/Lab/Excel_validation_logic.ipynb
```diff
@@ -0,0 +1,827 @@
...
+    "            else:\n",
+    "                e=0 \n",
+    "            \n",
+    "        if 'unknown' in df.iloc[l]['location_type'].lower():\n",
+    "            if not pd.isna(df.iloc[l]['address']) or not pd.isna(df.iloc[l]['latitude']) or not pd.isna(df.iloc[l]['longitude']):\n",
+    "                print(f'LOCATION ERROR ON ENTRY {l+1}: address/latitude/longitude not allowed')\n",
```
elpamart, line 562: same as above

elpamart (changes requested):
Some minor changes. But in general looks good to me :) I guess at some point we have to consider a validation to check that the data ingested exist in the ddbb apart from the pre-validation, but that is something we can improve in the future.

AngelArcones:
Updated with the changes and comments

elpamart (approved):

PR #238, opened 2022-02-03
scroll to scenarios list

mluena:
**DESCRIPTION**

This PR allows scrolling over scenarios list preventing the scroll of the whole sidebar. The "Create a new scenario" button and filters on top should be visible all time.

** JIRA**
https://vizzuality.atlassian.net/browse/LANDGRIF-155?atlOrigin=eyJpIjoiYWI2OTY1MDRhNmNkNGMyNzk4OTVlYzE3ZjY2ZWRiMjciLCJwIjoiaiJ9

** TESTING**


davidsingal:
image

Here you can see how it looks in Chrome for Windows. It may need a overflow hidden or something like that

davidsingal:
Also, if you can confirm with Andreia how it should be. Because if I'm not wrong the initial idea was to have an infinite scroll, so instead of load everything you have to load first some of them, and when you scroll load more using the pagination in the API.

PR #240, opened 2022-02-04
Add imported materials list endpoint

yulia-bel:
Endpoint within Materials module which gets a list of materials, imported by User with the related information from Sourcing locations (suppliers, location types, business units, etc)


api/test/e2e/materials/materials-get-list.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  afterEach(async () => {
+    await materialRepository.delete({});
+  });
+
+  afterAll(async () => {
+    await app.close();
+  });
+
+  test('Getting list of materials uploaded by users (and being part of sourcing locations) should be successful with default pagination', async () => {
```
tiagojsag, line 44: A few things I think can be improved in this test: 
- Add a few more materials that are not related to all the other entities, so you can validate that the endpoint filters them out correctly
- I am ok with two requests on the same test, but order them like "request1" -> "assertions1" -> "request2" -> "assertions2" as it make it easier to read.
- Your tests assume an order in the response materials ([bananas, cotton, maize]) that is not enforced anywhere. This is a nice way to get future false negatives
- You are repeating literals, like "bananas" or "maize" in your code. Those should only appear once, and then you reuse them by reference, so that it's easier to modify going forward (nitpick)

api/test/e2e/materials/materials-get-list.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+      .expect(HttpStatus.OK);
+
+    const responseWithCustomPagination = await request(app.getHttpServer())
+      .get(`/api/v1/materials/materials-list?page[size]=2`)
+      .set('Authorization', `Bearer ${jwtToken}`)
+      .send()
+      .expect(HttpStatus.OK);
+
+    expect(responseWithDefaultPagination.body.data[0].attributes.name).toEqual(
+      'bananas',
```
tiagojsag, line 84:
```suggestion
      material1.name,
```

(or whatever)

api/test/e2e/materials/materials-get-list.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    });
+
+    const responseWithDefaultPagination = await request(app.getHttpServer())
+      .get(`/api/v1/materials/materials-list`)
+      .set('Authorization', `Bearer ${jwtToken}`)
+      .send()
+      .expect(HttpStatus.OK);
+
+    const responseWithCustomPagination = await request(app.getHttpServer())
+      .get(`/api/v1/materials/materials-list?page[size]=2`)
```
tiagojsag, line 78:
```suggestion
      .get(`/api/v1/materials/materials-list`)
      .query({'page[size]': 2})
```

tiagojsag (changes requested):

yulia-bel:
@tiagojsag thanks a lot for the review - most of the issues are fixed. 

Regarding the endpoint, since now it is part of the sourcing locations module, it will be '/sourcing-locations/materials-list'.
In my opinion it makes sense to have this endpoint in sourcing-locations not only to take advantage of the nested relations and simplify the query, but also from the business logic point of view: this materials list is basically a reflection of sourcing locations and sourcing records of the user, but with the focus on materials in it's structure. @alexeh please, correct me if I am wrong, since you have more profound understanding of that :)

Tests refactored as well, the only comment regarding the order - the query has 'order by material name', to get them in alphabetical order, so in the test I wanted to cover that part as well

api/src/modules/sourcing-locations/sourcing-location.repository.ts
```diff
@@ -1,5 +1,31 @@
...
 import { SourcingLocation } from 'modules/sourcing-locations/sourcing-location.entity';
 
 @EntityRepository(SourcingLocation)
-export class SourcingLocationRepository extends Repository<SourcingLocation> {}
+export class SourcingLocationRepository extends Repository<SourcingLocation> {
+  async getSourcingLocationsMaterialsQuery(): Promise<
+    SelectQueryBuilder<SourcingLocation>
+  > {
+    const materialsListQuery: any = this.createQueryBuilder('sl')
+      .select([
```
alexeh, line 10: Didn't know this could be done passing the data in a array, nice!!

api/src/modules/sourcing-locations/materials-from-sourcing-locations.service.ts
```diff
@@ -0,0 +1,94 @@
...
+      obj.t1Supplier = obj.t1Supplier ? obj.t1Supplier.name : null;
+      obj.businessUnit = obj.businessUnit ? obj.businessUnit.name : null;
+
+      const purchases: { year: number; tonnage: number }[] = [];
+      for (const sr of obj.sourcingRecords) {
+        purchases.push({ year: sr.year, tonnage: sr.tonnage });
+      }
+      obj.purchases = purchases;
+      delete obj.sourcingRecords;
+    }
```
alexeh, line 90: Should we apply immutability here?

Or add it as tech debt and focus on progress?
@tiagojsag

api/src/utils/app-base.service.ts
```diff
@@ -88,7 +88,7 @@ export abstract class AppBaseService<
       this.pluralAlias,
       {
         ...this.serializerConfig,
-        paginationMeta,
+        meta: paginationMeta,
```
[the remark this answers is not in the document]
alexeh: This was the reason we weren't retrieving pagination data on the response. 
This change propagate said capabilities using the base service

Event tho according to typing this should accept a [key string] type, any prop that does not match 'meta' is being ignored by the serializer

Marxan has the same config this PR would include

api/src/modules/materials/materials.controller.ts
```diff
@@ -101,6 +106,30 @@ export class MaterialsController {
     return this.materialsService.serialize(results);
   }
 
+  @ApiOperation({
+    description:
+      'Get detailed list of materials imported by user and  existing in sourcing records',
+  })
+  @JSONAPIQueryParamsOnlyPagination()
+  @ApiOkResponse({ type: ImportedMaterialsListResponse })
+  @Get('/materials-list')
```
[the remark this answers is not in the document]
alexeh: Good point. We can talk about this in our next LG BE meet

api/src/modules/sourcing-locations/materials-from-sourcing-locations.service.ts
```diff
@@ -0,0 +1,94 @@
+import { Injectable, NotFoundException } from '@nestjs/common';
```
tiagojsag, line 1: not sure we need another service for this. what's the added value vs using the existing one?

api/src/modules/sourcing-locations/serializer/sourcing-locations-materials.serializer.ts
```diff
@@ -0,0 +1,69 @@
...
+      t1Supplier: sourcingLocation.t1Supplier?.name || null,
+      businessUnit: sourcingLocation.businessUnit?.name,
+      country: sourcingLocation.locationCountryInput,
+      locationType: sourcingLocation.locationType,
+      purchases: sourcingLocation.sourcingRecords.map((sr: SourcingRecord) => ({
+        year: sr.year,
+        tonnage: sr.tonnage,
+      })),
+    };
+
```
alexeh, line 66: Nice!

alexeh (approved):

PR #241, opened 2022-02-07
Feature LANDGRIF-388 Return user data for impact features

alexeh:
This PR adds: New endpoints to retrieve Admin-Regions, Materials, Suppliers... imported by a user

Deleted depth query param from the endpoints as we can't guarantee user stuff being within the client selected depth for the tree

I tried different approaches and I came up to realise the most performant way being:

-Select relevant item ID's from sourcing locations (not the entities)
-Get all data from the required table
-Filter full data getting relevant elements: the entities itself and all their ancestry
-Recursively build a tree

NOTE: This PRs include some fixes/improvements regarding the geocoding process


api/src/modules/admin-regions/admin-region.repository.ts
```diff
@@ -60,17 +65,27 @@ export class AdminRegionRepository extends ExtendedTreeRepository<
           st_setsrid($1::geometry, 4326),
           st_setsrid(g."theGeom"::geometry, 4326)
           )
-        AND ST_DWithin(g."theGeom" , st_setsrid($1::geometry, 4326), 1000)
-        AND a.level IS NOT NULL
         ORDER BY a.level DESC
-        LIMIT 1`,
+        LIMIT 3`,
```
tiagojsag, line 69: shouldn't the "order by a.level" make this and the "reduce()" dedundant? can you give me more detail as to what exactly happens to make this necesary?

api/src/modules/impact/impact.controller.ts
```diff
@@ -24,4 +27,36 @@ export class ImpactController {
...
+    description:
+      'Return admin-regions where a user has sourcing-locations within',
+  })
+  @ApiOkResponse({
+    type: AdminRegion,
+    isArray: true,
+  })
+  @Get('admin-regions')
+  async getAdminRegionTreeForImpact(): Promise<AdminRegion[]> {
+    return this.impactService.getAdminRegionTreeForImpact();
```
tiagojsag, line 51: serialization should go here, not on the service method

api/src/modules/impact/impact.controller.ts
```diff
@@ -24,4 +27,36 @@ export class ImpactController {
...
   }
+
+  @ApiOperation({ description: 'Return materials loaded by a user' })
+  @ApiOkResponse({
+    type: Material,
+    isArray: true,
+  })
+  @Get('materials')
+  async getMaterialTreeForImpact(): Promise<Material[]> {
+    return this.impactService.getMaterialTreeForImpact();
```
tiagojsag, line 38: so the impact module has this controller and this endpoint, which is named after material, but then the service uses the sourcing location service, and the material service, and then it all is serialized through a material module serializer.... not sure I like it...

api/src/modules/impact/impact.service.ts
```diff
@@ -163,4 +173,101 @@ export class ImpactService {
...
+    const supplierLineage: Supplier[] = this.getAncestry<Supplier>(
+      allSuppliers,
+      supplierIdsFromSourcingLocations,
+    );
+    return this.suppliersService.serialize(
+      this.buildTree<Supplier>(supplierLineage, null),
+    );
+  }
+
+  getAncestry<T extends { id: string; parentId?: string; children?: T[] }>(
```
tiagojsag, line 235: please document what this does. I tried looking at it, but my brain started to hurt. I need to re-review this once docs are in place

api/src/modules/materials/materials.service.ts
```diff
@@ -21,6 +22,7 @@ export class MaterialsService extends AppBaseService<
   constructor(
     @InjectRepository(MaterialRepository)
     protected readonly materialRepository: MaterialRepository,
+    protected readonly sourcingLocationsService: SourcingLocationsService,
```
tiagojsag, line 25: is this needed?

tiagojsag (changes requested):
I have the same feeling here that I had with @yulia-bel PR: we are adding new endpoints to basically load the same data, in a very similar way, just with an additional filter. I cna't help but wonder if all of this couldn't be done by tweaking existing code,rather than adding so much new code.

api/src/modules/admin-regions/admin-region.repository.ts
alexeh: There are 2 functions to determine de admin-region given some coordinates. In the use-case of this method we can't determine the granularity of the intersection, it might end-up returning a intersection which it's level might be null (or not)

That's why I decided to retrieve the final value within the API and make sure PostGIS retrieves some intersections.
Which would be a better approach?

api/src/modules/materials/materials.service.ts
alexeh: No it's not, good catch thanks!

alexeh:
@tiagojsag Refactored everything so that each entity controller retrieves it's data but related to ingested stuff

new endpoints will match

/api/v1/entityName/trees/user

This could also scale to a role based filter if needed in the future.
Thanks for the suggestion, I think it's better now

PR #242, opened 2022-02-07
qa indicator record calculation with updated methodology

elpamart:
The notebook contains an updated version of the queries to compute the indicator record values. This queries just contain the average version rather than the extended one with the probability purchase area as this would be more complex in terms of performance.


ikerey:
water impact (m3/yr) = (BWF * 0.001 / Prod all crops) * (ha / total ha) * Volume

This `Prod all crops`  has to bee replaced by `Prod all  **irrigated** crops`.

fgassert:
Hey,

I made some progress on this. There's still more that I'd like to review and confirm, but I wanted to share the first bit of what I was able to do. 

1. It looks like the reason why the calculations are slow is that the SQL is executing the same computation for each hex in the georegion. This means that for a georegion with 100 hexes, its calculating the sums 100 times, dividing them, and then adding them back together.
2. I'm wondering if we should try writing the equations so that its clear what has to be computed for each hex, and what should be computed using the sums across the whole georegion.

I've added another notebook with some rewritten SQL functions for you to review. It executes very fast, maybe 10-15 seconds for all records. Let me know if you'd like to review together and work from there.

Thanks!

elpamart:
CLARIFICATION NOTE: The water impact calculation formula is just a quick solution for getting the average blue water footprint for all the crops (just considering the mapspam datasets for a simple solution that doesn't involve any complex logic on the BE). As this methodology has to evolve in order to take into account [crop related blue water footprints](https://www.waterfootprintassessmenttool.org/world/accounting/table/blue) and this will lead to data models updates, we are not considering any change in the data model at this point until we don’t have a clear approach for this water impact calculations.

elpamart:
@fgassert thanks for your work on this! I don't fully understand why the SQL is executing the same computation for each hex in the georegion but I have taken a look to the formulas that you are applying and I have some questions. Please let me know when it's a good moment to look at this together.

PR #243, opened 2022-02-08
Tables styling according to designs + base admin/target UI

SARodrigues:
## Description

**In this PR:**

- New `Table` container component  
    - Custom styling matching the designs embedded in the component
        _To simplify the usage of tables and allow tables throughout the app to have a consistent design/spacing/colours. The new component abstracts most of the logic related to custom code, moving it to custom components that can be easily imported. _
    - `elementAttributes` and `content` can be overridden if needed, to maintain some flexibility    
    - Prop to make the first column sticky  
        _Enabled by default, will automatically set the shadow on the sticky column as per designs_  
    - Custom `SummaryRow` component  
        _To be used as an already styled row for summaries_  
    - Custom `GroupRow` component  
        _To help with expanding/collapsing_ grouped rows   
        - Can be further customised  
        - The whole row is clickable in order to `expand`/`collapse` grouped rows  
     - Custom cell `types`:  
        - `LineChartCell`: making it easy to display a (styled) line chart as a table cell  
- Removed the no longer needed original generic `Table` component, to prevent confusion  
- `Analysis` section tables now use the new `Table` component  
    _Code logic has not been changed. A few tweaks were made just to pass the correct data to make the new component work_  
- `Admin` section tables now use the new `Table` component  
    _Code logic has not been changed. A few tweaks were made just to pass the correct data to make the new component work_  
- Several spacing/color tweaks to better match the designs  
- The `Materials` tab in the admin area was renamed to `Data`, as it better matches the designs (and makes more sense IMHO)  

**NOTES:**
- Quite a few tweaks will most likely be needed along the way to make it work with future/in progress design changes, however it is useable and hopefully lays the groundwork for an easily modifiable component with custom cell/row types. It should also abstract quite a bit of design/custom cell logic.   
- The `MultipleIndicatorTable` & `SingleIndicatorTable` components share quite a bit of code, which was repeated. I abstracted from changing the logic so that the PR would focus on the tables. It is something to keep in mind for the future though; those components can be further simplified and possibly merged in the future.  

## JIRA  
[LANDGRIF-490 (Style tables according to design)](https://vizzuality.atlassian.net/browse/LANDGRIF-490)
[LANDGRIF-463 (Add UI for target management)](https://vizzuality.atlassian.net/browse/LANDGRIF-463)

## Screenshots  

**Analysis**  
Screenshot 2022-02-08 at 15 31 39

**Admin/targets**  
Screenshot 2022-02-08 at 15 32 30

## Videos    

https://user-images.githubusercontent.com/6273795/153019895-03d07ad0-d919-4e7d-a3c0-08f9f4875771.mp4


davidsingal (approved):
Good job!

PR #245, opened 2022-02-10
Add search and orderBy flags to SL materials endpoint

yulia-bel:


api/src/decorators/json-api-parameters.decorator.ts
```diff
@@ -157,3 +157,19 @@ export const JSONAPISingleEntityQueryParams = (fetchConfiguration?: {
     fieldsQueryParam,
     omitFieldsQueryParam,
   );
+
+export const JSONAPISourcingMaterialsQueryParams = (fetchConfiguration?: {
```
tiagojsag, line 161: This seems to be a subset of  `JSONAPIQueryParams`, removing a few decorators that seem to do nothing by default, so it's not clear why we need it. Also, even if we do, the name is domain-specific, so if we do decide to keep it, it needs to be renamed

api/src/modules/sourcing-locations/sourcing-locations.service.ts
```diff
@@ -90,24 +90,40 @@ export class SourcingLocationsService extends AppBaseService<
...
+
+    if (fetchSpecification.orderBy === 'country') {
+      query.orderBy(`${this.alias}.locationCountryInput`);
+    } else if (fetchSpecification.orderBy === 'locationType') {
+      query.orderBy(`${this.alias}.locationType`);
+    } else {
+      fetchSpecification.orderBy
+        ? query.orderBy(`${fetchSpecification.orderBy}.name`)
+        : query.orderBy('material.name');
+    }
```
tiagojsag, line 124: the trenary operator in the `else` clause is evil because it's effectively a 3rd `if` clause, in disguise. And 3 `if` clauses in a row is a great reason to refactor this using a `switch` instead

api/src/modules/sourcing-locations/dto/materials-query.sourcing-location.dto.ts
```diff
@@ -0,0 +1,30 @@
...
+  LOCATION_TYPE = 'locationType',
+}
+
+export class GetSourcingMaterialsQueryDto {
+  @ApiProperty({ required: false, enum: MATERIALS_ORDER_BY_VALUES })
+  @IsOptional()
+  @IsEnum(MATERIALS_ORDER_BY_VALUES, {
+    message:
+      'Available columns for orderBy: country, material, producer, t1Supplier, locationType',
+  })
```
tiagojsag, line 19:
```suggestion
  @IsEnum(MATERIALS_ORDER_BY_VALUES, {
    message: `Available columns for orderBy: ${Object.values(
      MATERIALS_ORDER_BY_VALUES,
    ).join(', ')}`,
  })
```

tiagojsag: Also: your enum has 6 values, but the error message only has 5 - which is the perfect example of why my change suggestion makes sense. tests fail once you change it, so mind that too

api/src/modules/sourcing-locations/sourcing-locations.controller.ts
```diff
@@ -38,6 +38,7 @@ import { UpdateSourcingLocationDto } from 'modules/sourcing-locations/dto/update
 import { PaginationMeta } from 'utils/app-base.service';
 import { SourcingLocationsMaterialsResponseDto } from 'modules/sourcing-locations/dto/materials.sourcing-location.dto';
 import { SourcingLocationsMaterialsSerializer } from 'modules/sourcing-locations/serializer/sourcing-locations-materials.serializer';
+import { GetSourcingMaterialsQueryDto } from './dto/materials-query.sourcing-location.dto';
```
tiagojsag, line 41: this relative import should cause ESLint to report a problem. Please check your IDE settings, and use an absolute import instead

tiagojsag (changes requested):

api/src/decorators/json-api-parameters.decorator.ts
yulia-bel: Here I tried to exclude the query params which may have unexpected behavior when endpoint returns materials with 'materialsData=true' (such as omit, sort, field). Those flags come from base-service and work with the columns of the original entity table (sourcing_locations), but not with the joined ones (for example, we cannot filter by material name, because 'material' comes as an array of joined relations, it is not a column of 'sourcing_locations' table).
However, since JSONAPIQueryParams is still correct for this endpoint without 'materialsData=true', I would probably leave it and try to document the above mentioned details..

api/src/modules/sourcing-locations/dto/materials-query.sourcing-location.dto.ts
```diff
@@ -0,0 +1,31 @@
...
+  MATERIAL = 'material',
+  LOCATION_TYPE = 'locationType',
+}
+
+export class GetSourcingMaterialsQueryDto {
+  @ApiProperty({ required: false })
+  @IsOptional()
+  materialsData?: boolean;
+
+  @ApiProperty({ required: false, enum: MATERIALS_ORDER_BY_VALUES })
```
alexeh, line 18: Shouldn't this be 
`@ApiPropertyOptional()`?

api/src/modules/sourcing-locations/dto/materials-query.sourcing-location.dto.ts
```diff
@@ -0,0 +1,31 @@
...
+  @IsOptional()
+  @IsEnum(MATERIALS_ORDER_BY_VALUES, {
+    message: `Available columns for orderBy: ${Object.values(
+      MATERIALS_ORDER_BY_VALUES,
+    ).join(', ')}`,
+  })
+  orderBy?: MATERIALS_ORDER_BY_VALUES;
+
+  @ApiProperty({ required: false })
```
alexeh, line 27: Same for this

alexeh (approved):
Thanks @yulia-bel nice job! I just left 2 minor improvement suggestions

Feel free to merge with or without applying them

api/src/decorators/json-api-parameters.decorator.ts
tiagojsag: kudos for testing out those scenarios ;)

ok, but taking your example, and since this endpoint would return SLs, not Materials, the "omit" would only apply to the SL fields, not the Material fields, right?

If you still think the slimmed down version of the decorator is the best option, the go for it, but use a different name for it, so it describes the function of the decorator, not where it's used.

PR #246, opened 2022-02-10
Feature LANDGRIGF-503 endpoints for interventions suppliers and lg stimates

alexeh:
This PR adds:

First piece of: Endpoints to retrieve suppliers by type, List Business Units with sourcing locations in a tree format

Adds a lot of performance optimisation to previous implementation


api/src/modules/admin-regions/admin-region.repository.ts
```diff
@@ -88,4 +89,52 @@ export class AdminRegionRepository extends ExtendedTreeRepository<
       return previous.level > current.level ? previous : current;
     });
   }
+
+  /**
+   * @description Retrieves admin-regions and it's ancestors (in a plain format) there are registered sourcingLocations within
```
tiagojsag, line 94:
```suggestion
   * @description Retrieves admin regions and their ancestors (in a plain format) there are registered sourcingLocations within
```

api/src/modules/admin-regions/admin-region.repository.ts
```diff
@@ -88,4 +89,52 @@ export class AdminRegionRepository extends ExtendedTreeRepository<
...
+            join adminregion_tree p on p."parentId" = c.id
+        )
+        select distinct *
+        from adminregion_tree
+        order by name`,
+
+      subQueryParams,
+    ).catch((err: Error) =>
+      this.logger.error(
+        `Query Failed for retrieving admin-regions with sourcing locations: `,
```
tiagojsag, line 128:
```suggestion
        `Query failed for retrieving admin regions with sourcing locations: `,
```

api/src/modules/admin-regions/admin-region.repository.ts
```diff
@@ -88,4 +89,52 @@ export class AdminRegionRepository extends ExtendedTreeRepository<
...
+        err,
+      ),
+    );
+
+    if (!result || !result.length)
+      throw new NotFoundException(
+        'No Admin Regions with sourcing locations within found. Please check if sourcing-data has been provided to the platform',
```
tiagojsag, line 135:
```suggestion
        'No Admin Regions with associated Sourcing Locations found. Please check if sourcing data has been provided to the platform',
```

api/src/modules/business-units/business-unit.repository.ts
```diff
@@ -1,10 +1,61 @@
...
+  logger: Logger = new Logger(BusinessUnitRepository.name);
+
+  /**
+   * @description Retrieves business-units and it's ancestors (in a plain format) there are registered sourcingLocations for
+   */
+
+  async getSourcingDataAdminRegionsWithAncestry(): Promise<BusinessUnit[]> {
+    // Join and filters over business-units present in sourcing-locations. Resultant query returns IDs of elements meeting the filters
+    const queryBuilder: SelectQueryBuilder<BusinessUnit> =
+      this.createQueryBuilder('bu')
```
tiagojsag, line 22: Couldn't you just "select distinct sourcingLocation.businessUnitId from sourcingLocation" and avoid the join?

api/src/modules/business-units/business-unit.repository.ts
```diff
@@ -1,10 +1,61 @@
...
+      queryBuilder.getQueryAndParameters();
+
+    // Recursively find elements and their ancestry given Ids of the subquery above
+    const result: BusinessUnit[] = await this.query(
+      `
+        with recursive businessunit_tree as (
+            select m.id, m."parentId", m."name", m."description"
+            from business_unit m
+            where id in
+                        (${subQuery})
```
tiagojsag, line 37: any way we could to without the subquery? Typically they are a performance sink

api/src/modules/materials/material.repository.ts
```diff
@@ -1,10 +1,62 @@
...
+   * @description Retrieves materials and it's ancestors (in a plain format) there are registered sourcingLocations for
+   */
+
+  async getSourcingDataMaterialsWithAncestry(): Promise<Material[]> {
+    // Join and filters over materials present in sourcing-locations. Resultant query returns IDs of elements meeting the filters
+    const queryBuilder: SelectQueryBuilder<Material> = this.createQueryBuilder(
+      'm',
+    )
+      .select('m.id')
+      .innerJoin(SourcingLocation, 'sl', 'sl.materialId = m.id')
```
tiagojsag, line 25: same potential query simplification as above

api/src/modules/materials/material.repository.ts
```diff
@@ -1,10 +1,62 @@
...
+      queryBuilder.getQueryAndParameters();
+
+    // Recursively find elements and their ancestry given Ids of the subquery above
+    const result: Material[] = await this.query(
+      `
+        with recursive material_tree as (
+            select m.id, m."parentId", m."name", m.description, m."createdAt", m."updatedAt", m."hsCodeId"
+            from material m
+            where id in
+                        (${subQuery})
```
tiagojsag, line 38: and here too

api/src/modules/materials/material.repository.ts
```diff
@@ -1,10 +1,62 @@
...
+            union all
+            select c.id, c."parentId", c."name", c.description, c."createdAt", c."updatedAt", c."hsCodeId"
+            from material c
+            join material_tree p on p."parentId" = c.id
+        )
+        select distinct *
+        from material_tree
+        order by name`,
```
tiagojsag, line 46: do we need the order by? if we have no need to sort results, remove it, as it can be a performance sink

api/src/modules/suppliers/dto/get-supplier-by-type.dto.ts
```diff
@@ -0,0 +1,14 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsEnum, IsNotEmpty } from 'class-validator';
+import { SUPPLIER_TYPES } from 'modules/suppliers/supplier.entity';
+
+export class GetSupplierByType {
+  @ApiProperty()
+  @IsNotEmpty()
+  @IsEnum(SUPPLIER_TYPES, {
+    message: `Available Supplier types: ${Object.values(SUPPLIER_TYPES).join(
```
tiagojsag, line 9:
```suggestion
    message: `Allowed Supplier types: ${Object.values(SUPPLIER_TYPES).join(
```

api/src/modules/suppliers/suppliers.service.ts
```diff
@@ -101,26 +117,29 @@ export class SuppliersService extends AppBaseService<
 
   /**
    *
-   * @description Get a tree of Suppliers imported by a User
+   * @description Get a tree of Suppliers with sourcing locations
```
tiagojsag, line 120:
```suggestion
   * @description Get a tree of Suppliers that are associated with sourcing locations
```

api/src/modules/materials/material.repository.ts
```diff
@@ -1,10 +1,62 @@
...
+
+  async getSourcingDataMaterialsWithAncestry(): Promise<Material[]> {
+    // Join and filters over materials present in sourcing-locations. Resultant query returns IDs of elements meeting the filters
+    const queryBuilder: SelectQueryBuilder<Material> = this.createQueryBuilder(
+      'm',
+    )
+      .select('m.id')
+      .innerJoin(SourcingLocation, 'sl', 'sl.materialId = m.id')
+      .distinct(true);
+
```
alexeh, line 27: In case it's not clear, I am not retrieving the query and params straight off the chain of createQueryBuilder because we will need to add additional filters here in the future, so we easily do this like:

if(condition)
  queryBuilder.andWhere(...)

api/test/e2e/admin-regions/admin-regions-get-trees.spec.ts
```diff
@@ -263,10 +263,10 @@ describe('AdminRegions - Get trees', () => {
         .set('Authorization', `Bearer ${jwtToken}`)
         .expect(HttpStatus.OK);
 
-      expect(response.body.data[0].id).toEqual(parentWithNoChildAdminRegion.id);
-      expect(response.body.data[0].attributes.children).toEqual([]);
-      expect(response.body.data[1].id).toEqual(parentAdminRegion.id);
-      expect(response.body.data[1].attributes.children[0].id).toEqual(
+      expect(response.body.data[1].id).toEqual(parentWithNoChildAdminRegion.id);
```
tiagojsag, line 266: this feels like a test depending on a specific order of elements, which typically it should not unless the api endpoint makes a commitment around the order in which it returns the results

api/test/e2e/suppliers/suppliers-get-by-type.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      async () => {
+        const response = await request(app.getHttpServer())
+          .get(`/api/v1/suppliers/types`)
+          .query({ type: 'I have not sell anything in my life' })
+          .set('Authorization', `Bearer ${jwtToken}`)
+          .send()
+          .expect(HttpStatus.BAD_REQUEST);
+        expect(
+          response.body.errors[0].meta.rawError.response.message[0],
+        ).toEqual('Available Supplier types: t1supplier, producer');
```
tiagojsag, line 93:
```suggestion
        const response = await request(app.getHttpServer())
          .get(`/api/v1/suppliers/types`)
          .query({ type: 'I have not sell anything in my life' })
          .set('Authorization', `Bearer ${jwtToken}`)
          .send();
          
        expect(response.<wahtever>).toEqual(HttpStatus.BAD_REQUEST);
        expect(
          response.body.errors[0].meta.rawError.response.message[0],
        ).toEqual('Available Supplier types: t1supplier, producer');
```

review the actual implementation

tiagojsag (commented):
I left a few comments that I think apply to multiple parts of your code, since it seems you are doing the same thing in multiple places for different entities.

Also: please please, make sure entity names are consistent. My OCD is killing me

api/src/modules/business-units/business-unit.repository.ts
alexeh: Yes I could, the thing is this is split by entities so each repository points to its target entity by default, so I need either add services in Sourcing Location services to retrieve required Ids for different entities, went for this option to keep stuff organised

Or import the standalone manager in each entity with seems nasty as well, at the end of the day I am just retrieving Ids, nothing else from the DB

The trade-off between performance and developer experiences seems good to me. Having said that, let me know if you still want to move this

api/src/modules/materials/material.repository.ts
alexeh: I guess we need the orderBy to enforce the lists to retrieve in the same order (this is for a UI dropdown)

api/test/e2e/admin-regions/admin-regions-get-trees.spec.ts
alexeh: Yes, this is enforced by the orderBy you pointed out above

PR #247, opened 2022-02-11
Adding search & years range filters to Admin/data

SARodrigues:
## Description

This PR adds searching and years filtering to the admin/data page. In addition, it simplifies using the range filter throughout the application, including in the Analysis view.  

**In this PR:**
- New `Search` (input) component  
  _Styled to match the app_  
- New `YearsRangeFilter` component  
    _To promote reusability of the filter button throughout the app._   
    - `useYearsRange` hook
      _Helper hook specific to the `YearsRangeFilter` component, to simplify the usage of the filter and centralise the code specific to years range filtering, as well as validating initialisations_   
- New `NoResults` component  
    _Primitive generic component currently in use in the admin views, to display when there are no results stemming from searching/filtering data_
- `Table` component changes:  
    - Small styling tweaks to better match the 
- `Select` component changes:  
    - Fixed an issue where the cursor would disappear while hovering a disabled list item
- `Admin/data` tab chances:  
    - Added search through the API endpoint  
    - Added frontend years filtering  
- `Analysis` tab changes  
    - `AnalysisFilters` `Years` filter now makes use of the `YearsRangeFilter` component  
        _Only for years ranges_  
    - Clicking outside the filter dropdown now closes it  
        _Matching other dropdowns we have in the app_

## JIRA  
[LANDGRIF-506 (Admin/Data table should be able to be searched by text)](https://vizzuality.atlassian.net/browse/LANDGRIF-506)  
[LANDGRIF-512 (Add years filters to Admin/Data)](https://vizzuality.atlassian.net/browse/LANDGRIF-512)

## Screenshots
Screenshot 2022-02-14 at 15 53 51
Screenshot 2022-02-14 at 15 54 18

### DIFF
--- a/client/src/containers/analysis-visualization/analysis-filters/years/component.tsx
+++ b/client/src/containers/analysis-visualization/analysis-filters/years/component.tsx
@@ -1,4 +1,5 @@
-import React, { useEffect, useMemo } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import { isFinite, toNumber } from 'lodash';
 
 import { useAppDispatch, useAppSelector } from 'store/hooks';
 import { analysis, setFilter } from 'store/features/analysis';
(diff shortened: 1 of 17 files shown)


client/src/containers/analysis-visualization/analysis-filters/years/component.tsx
```diff
@@ -1,224 +1,59 @@
...
   const { visualizationMode, filters, layer } = useAppSelector(analysis);
-  const { startYear, endYear, materials, indicator } = filters;
-  const [additionalYear, setAdditionalYear] = useState<number>(null);
-  const { data, isLoading } = useYears(layer, materials, indicator);
-  const dataWithAdditionalYear: number[] = useMemo(() => {
-    const result = [...(data || [])];
-    if (additionalYear) result.push(additionalYear);
-    return result;
-  }, [data, additionalYear]);
-  const handleAdditionalYear: SelectProps['onSearch'] = useCallback(
```
SARodrigues, line 27: I didn't include this, as I didn't understand what it does. What is the goal of the additional year?

davidsingal: We want to allow the user to choose a year that is not in the database because the app is able to calculate new years. But this is only needed in Analysis. For data and admin purposes it doesn't make any sense.

davidsingal: So, we will need it but only for Analysis. Could you isolate this functionality to be applied only in Analysis?

davidsingal (changes requested):

client/src/containers/analysis-visualization/analysis-filters/years/component.tsx
SARodrigues: @davidsingal I've restored the behavior :)

davidsingal (approved):

PR #248, opened 2022-02-11
Add order param to SL materials endpoint

yulia-bel:


api/src/modules/sourcing-locations/dto/materials-query.sourcing-location.dto.ts
```diff
@@ -20,6 +25,15 @@ export class GetSourcingMaterialsQueryDto {
...
   orderBy?: MATERIALS_ORDER_BY_VALUES;
 
+  @ApiPropertyOptional({ enum: MATERIALS_ORDER_BY_OPTIONS })
+  @IsOptional()
+  @IsEnum(MATERIALS_ORDER_BY_OPTIONS, {
+    message: `Available columns for order: ${Object.values(
+      MATERIALS_ORDER_BY_OPTIONS,
+    ).join(', ')}`,
+  })
+  order?: string;
```
alexeh, line 35:
```suggestion
  order?: MATERIALS_ORDER_BY_OPTIONS;
```

Also, I think this should be something more generic as DESC and ASC could be applied to almost everything. But no need to refactor it now, minor improvement for the future

alexeh (approved):
Thanks @yulia-bel, just a minor comment, please change it and feel free to merge!!

PR #250, opened 2022-02-15
Fixed 500 error when the user change of layer in analysis page

davidsingal:
https://vizzuality.atlassian.net/browse/LANDGRIF-515


alexeh (approved):
That was fast 😄 

Just checked the preview, looks good. Thank you @davidsingal !

PR #251, opened 2022-02-16
Landgrif 497 create an scenario tab

mluena:
First layout for New scenario and create an intervention.


davidsingal (approved):
There will be required more adjustments but they could be covered in other tasks

PR #256, opened 2022-02-18
Add `withSourcingLocations` query param to Analysis/Impact filters

SARodrigues:
## Description  

**In this PR:**
Added `withSourcingLocations` query param to Analysis/Impact filters, namely `Material`, `Origins` and  `Suppliers` filters in the `AdditionalFilters` component.  

**Notes:**  
1. The reason I opted for adding `depth` and `withSourcingLocations` as `props`, is because even though these filters are specific to the Analysis view, they are not specific to _Impact_. For instance, the `Materials` filter is also used when the user is on _Material production_ and not _Impact_.

    I feel like adding these `props` to one filter and not the others could become confusing in the future, so I added the same ones to all three. 

2. I made some alignment tweaks to the items in the `TreeSelect` component. Reason being, since with these filters some top level items may not have children, the expanding arrow would not appear, and it was confusing to see which items/children were under which top level item. In order to fix that issue I added the equivalent space of the arrow to preserve the original spacing of the component.

## JIRA  
[LANDGRIF-520 (Add withSourcingLocations query param for Impact filters)](https://vizzuality.atlassian.net/browse/LANDGRIF-520)

## Screenshots

**Materials**
Screenshot 2022-02-18 at 15 30 02

**Origins** 
Screenshot 2022-02-18 at 15 30 10

**Suppliers** 
Screenshot 2022-02-18 at 15 30 14

### DIFF
--- a/client/src/components/tree-select/component.tsx
+++ b/client/src/components/tree-select/component.tsx
@@ -62,7 +62,7 @@ const TreeSelect: React.FC<TreeSelectProps> = ({
   );
 
   const customSwitcherIcon = useCallback(({ isLeaf, expanded }) => {
-    if (isLeaf) return null;
+    if (isLeaf) return <span className="block w-4" />;
     if (expanded) return <ChevronDownIcon className="h-4 w-4" />;
     return <ChevronRightIcon className="h-4 w-4" />;
   }, []);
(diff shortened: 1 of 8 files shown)


alexeh (approved):
Thanks @SARodrigues! Just checked the preview version and looks like is working nicely!

Regarding the scenarios part, I think once you get there you can add the dependant dropdown filtering as well (it should work adding required IDs of elements)

davidsingal (approved):

PR #257, opened 2022-02-21
Landgrif 516 new scenario step should change depending on intervention type

mluena:
Analysis Page / create a new scenario
- Layout in step 2 while creating a new intervention should change depending on the intervention type selected in step 1.
- Different layouts have been splitter in components.
- Select in step one should be multiple (materials, suppliers, and sourcing regions already changed)
- The treeSelect component has been splited in two components depending on styling


client/src/components/badge/types.d.ts
```diff
@@ -4,4 +4,5 @@ export type BadgeProps = {
   className?: string;
   removable?: boolean;
   onClick?: (data: BadgeProps['data']) => void;
+  background?: boolean;
```
davidsingal, line 7: This is fine, but for the naming, I would prefer to be more explicit on the functionality instead of the appearance. To me in this case, the `background` property doesn't have any functionality and it's better to allow add `className` from outside and the dev can decide to extend the styles as needed. Another valid option it's to define themes or variants.

client/src/components/select/types.d.ts
```diff
@@ -15,6 +15,7 @@ export type SelectProps = {
   disabled?: boolean;
   loading?: boolean;
   label?: string;
+  id?: string;
```
davidsingal, line 18: I think we can extend the props from HTMLSelect, but you can leave as it is.

client/src/components/tree-select/component.tsx
```diff
@@ -201,76 +193,44 @@ const TreeSelect: React.FC<TreeSelectProps> = ({
...
-            {selected ? (
-              <span className="font-medium">{selected.label}</span>
-            ) : (
-              <span className="text-gray-300">{placeholder}</span>
-            )}
-          </span>
-          <Icon />
-        </button>
+      {theme === 'primary' && (
+        <TreeSelectButtonPrimary
```
davidsingal, line 208: It's they have different functionality or behavior they should be different components. In case the props, functionality, and everything is the same but only change the theme, it should be only controlled by `classnames`.

client/src/containers/interventions/new/component.tsx
```diff
@@ -56,12 +80,24 @@ const InterventionForm = () => {
...
-            <button
-              type="submit"
-              className="ml-3 inline-flex justify-center py-2 px-4 border border-transparent shadow-sm text-sm font-medium rounded-md text-white bg-green-700 hover:bg-green-800 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500"
-            >
-              Continue
-            </button>
+            {!!currentStep && !isFirstStep && (
+              <button
+                type="submit"
+                className="ml-3 inline-flex justify-center py-2 px-4 border border-transparent shadow-sm text-sm font-medium rounded-md text-white bg-green-700 hover:bg-green-800 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500"
```
davidsingal, line 86: This button should be a component that we can reuse due to the styles mainly, or at least the className should be exported to use it.

client/src/containers/interventions/new/step1/component.tsx
```diff
@@ -8,104 +8,65 @@ import { analysis, setFilter } from 'store/features/analysis';
...
-  const supliers = ['supplier1', 'supplier2', 'supplier3'];
-  const sourcingRegions = ['sourcingRegion1', 'sourcingRegion2', 'sourcingRegion3'];
-  const yearCompletions = [2001, 2015, 2020];
-  const interventionTypes = [
-    'Source from a new supplier or location',
-    'Change production efficiency',
-    'Switch to a new material',
-  ];
 
   // const { data: materials, isLoading: isLoadingMaterials } = useMaterials();
```
davidsingal, line 32: why did you comment these lines? The data should come from those hooks, also the hardcoded data

client/src/store/features/analysis/index.ts
```diff
@@ -24,6 +24,7 @@ export type AnalysisState = {
     materials: Option[];
     origins: Option[];
     suppliers: Option[];
+    interventionType: number;
```
davidsingal, line 27: we only have three interventions type, so in case you want to use numbers, it should be limited. Anyway, I would prefer to be more explicit and use strings (slugs).

davidsingal (changes requested):

PR #262, opened 2022-02-22
Api/data ingestion restore deleted data on failure

asuramaru1:
Run everything in one transaction when importing sourcing file , to rollback everything as it was before on error


tiagojsag:
@asuramaru1 rebase please

PR #263, opened 2022-02-22
validate import file before creating task

asuramaru1:
validate xlsx  file before creating task or returning response


api/src/modules/import-data/sourcing-data/sourcing-data-import.service.ts
```diff
@@ -244,4 +248,42 @@ export class SourcingDataImportService {
...
+        `File: ${filename} could not have been loaded. Please try again later or contact the administrator`,
+      );
+    }
+
+    /**
+     * Pass random id to create DTOs for validation
+     * @note: If we decide to path DTOs to task instead of path
+     * this has to change to generate valid id
+     */
+    const sourcingLocationGroupId: string =
```
tiagojsag, line 273: this is a hack. any chance we can make the sourcingLocationGroupId an optional argument on the function below (and its dependencies), and validate without it?

tiagojsag (changes requested):

api/src/modules/import-data/sourcing-data/sourcing-data-import.service.ts
alexeh: By model constraints, a SoucingLocation needs to have a SourcingLocationGroupId. But yes, we can avoid this hack passing the 'real' Id of the SourcingLocationGroup we are creating when running an import.

Do you want me to do it in this PR or do you want to see it done in the next PR implementing the 'real' validation?

tiagojsag: if you promise you won't forget, you can do it in the next PR. but because I don't trust that you won't forget (I trust you, not your memory), add a @todo note here before merging, just in case

api/src/modules/suppliers/supplier.repository.ts
```diff
@@ -4,7 +4,7 @@ import { ExtendedTreeRepository } from 'utils/tree.repository';
 import { CreateSupplierDto } from 'modules/suppliers/dto/create.supplier.dto';
 import { Logger, NotFoundException } from '@nestjs/common';
 import { SourcingLocation } from 'modules/sourcing-locations/sourcing-location.entity';
-import { GetSupplierTreeWithOptions } from './dto/get-supplier-tree-with-options.dto';
+import { GetSupplierTreeWithOptions } from 'modules/suppliers/dto/get-supplier-tree-with-options.dto';
```
alexeh, line 7: Nice catch :)

api/src/modules/import-data/import-data.service.ts
```diff
@@ -19,6 +19,15 @@ export class ImportDataService {
     private readonly tasksService: TasksService,
   ) {}
 
+  async ValidateAndLoadXlsxFile(
```
alexeh, line 22: little one: camelCase here?

api/src/modules/import-data/sourcing-data/validators/sourcing-data.class.validator.ts
```diff
@@ -0,0 +1,118 @@
...
+  IsString,
+  Min,
+  MinLength,
+  Validate,
+} from 'class-validator';
+import { LOCATION_TYPES } from 'modules/sourcing-locations/sourcing-location.entity';
+import { LocationAddressInputValidator } from 'modules/import-data/sourcing-data/validators/address-input.custom.validator';
+import { LocationLatitudeInputValidator } from 'modules/import-data/sourcing-data/validators/latitude-input.custom.validator';
+import { LocationLongitudeInputValidator } from 'modules/import-data/sourcing-data/validators/longitude-input.custom.validator';
+
```
alexeh, line 15: I believe this can be a bit dangerous as we get tied to the current year amount of the excel. I think it is OK to match the columns as we rely on these to build create entities, build realtions and a lot of stuff.

But the 'year columns' will 100% change, users will add more years, and most likely different users will have different years. I think (I might be wrong) that this approach will just match the current dataset

There is a sorting algorithm that finds columns including some year value and extracting both the year and value from there. I believe it would be a better idea to validate after the algorithm has dynamically sorted this out, and then recursively validate the values in these fields

Also, probably Science is validating that these columns hold a numeric value, but we should also guard the API against unexpected data.

PS: Don't trust people xD

api/src/modules/import-data/sourcing-data/dto-processor.service.ts
```diff
@@ -174,6 +182,8 @@ export class SourcingRecordsDtoProcessorService {
       sourcingData.push(sourcingLocationWithSourcingRecords as SourcingData);
     }
 
+    await this.validateCleanData(sourcingData);
```
alexeh, line 185: This is a very personal opinion, but I think it would be more readable if we take this out of the <cleanData> process

alexeh:
Nice job @yulia-bel !

I added some suggestions including one for performance improvements. I can't functional-test it locally as it breaks my current DB model constraints (as I updated some enums of the intervention workflow and stuff) 

I suggest to handle this comments (if you think these are actually possible to apply) and merge so we can test it with the huge dataset (if you haven't already)

Thanks!

PR #264, opened 2022-02-22
create new geoCoding base service that caches api calls

asuramaru1:
create new geoCoding service that caches api responses. ( since geocoding is done in async way , there may still be more that one call to the api when ingesting file untill cache is saved in redis)


tiagojsag:
I created https://github.com/Vizzuality/landgriffon/pull/269 which pick up on both the existing code structure and you extension of it, and refactors things so it's a bit more structured.

Also: some tests would be welcome.

alexeh:
Implementation present in DB extended and merged in: https://github.com/Vizzuality/landgriffon/pull/269

Closing this, thanks!!

PR #268, opened 2022-02-22
Enabling export/download of materials data in admin/data

SARodrigues:
## Description  

This PR makes it so that clicking the `Download` button in the Admin/Data view, a CSV download containing the respective materials data is triggered. 

**In this PR:**

- New `Downloader` component  
  _Meant as component to make exporting & downloading data easier. Currently it only supports `csv`, but can be expanded to support other export types, or actual files._
  _As its minimum, the only required prop is a url pointing at an endpoint with data, and it'll build a `csv` out of it in addition of triggering a download._
    - Reusable; can be used for other endpoints or even different types of data  
    - It makes use of [`react-csv`](https://github.com/react-csv/react-csv) for the heavy lifting  
    - `params` can be passed to the endpoint
        _Should we need to add filters, sorting, etc to the `csv` file_  
    - Support for custom headers to be used in the resulting `csv`
    - Support for a `transform` function
        _So that the `headers`/`data` to be used in the `csv` file can be manipulated/cleaned up._
- New `DownloadMaterialsDataButton` component  
  _Triggers the export/download of the admin materials data  by making use of the above `Downloader` component_ 
    - Contains the definition for the properties/headers to be exported in the CSV  
    - Makes use of the `Downloader`'s `transformer` callback to manipulate the data & headers  
    - Button shows a `Loading` spinner while the data is being fetched  
- Admin/data view chances:  
    - Now the `Download` works  
    - If the data fetching/download fails, it'll show an error message

## JIRA  

[LANDGRIF-508 (Admin/Data data should be exportable and downloadable )](https://vizzuality.atlassian.net/browse/LANDGRIF-508)

## Screenshots

**Resulting CSV** 
csv

**Uploading** 
uploading

**Error**
error

### DIFF
--- a/client/src/containers/admin/download-materials-data-button/component.tsx
+++ b/client/src/containers/admin/download-materials-data-button/component.tsx
@@ -1,0 +1,86 @@
+import { useState } from 'react';
+import { flatten, noop, pick, uniq } from 'lodash';
+
+import Downloader, { DownloaderHeadersType, DownloaderTransformProps } from 'containers/downloader';
+
+import Button from 'components/button';
+import Loading from 'components/loading';
+
+import { DownloadMaterialsDataButtonProps } from './types';
+
+const DEFAULT_HEADERS: DownloaderHeadersType[] = [
+  { key: 'materialName', label: 'Material' },
+  { key: 'businessUnit', label: 'Business unit' },
+  { key: 't1Supplier', label: 'Tier 1 Supplier' },
+  { key: 'producer', label: 'Producer' },
+  { key: 'locationType', label: 'Location type' },
+  { key: 'country', label: 'Country' },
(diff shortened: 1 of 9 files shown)


client/src/containers/admin/download-materials-data-button/component.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const handleOnError = () => {
+    setIsDownloading(false);
+    onError('There was an error processing the data for download.');
+  };
+
+  return (
+    <Downloader
+      className={className}
+      url="/sourcing-locations/materials"
```
SARodrigues, line 72: Didn't set a type since it defaults to `csv` (only format currently supported).

PR #278, opened 2022-02-28
fix: prevent dates comp re-render

mluena:
**DESCRIPTION**
This PR updates useEffect in the Select component to prevent an infinitive loop while setting the selected year

**JIRA**
https://vizzuality.atlassian.net/browse/LANDGRIF-535?atlOrigin=eyJpIjoiYjNjMDEzMWM4NDkwNDcyM2EyMzY3ODdjZDA1OGQ1NDEiLCJwIjoiaiJ9

### DIFF
--- a/client/src/components/select/component.tsx
+++ b/client/src/components/select/component.tsx
@@ -47,16 +47,13 @@ const Select: React.FC<SelectProps> = (props: SelectProps) => {
   const [searchTerm, setSearchTerm] = useState<string>('');
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


davidsingal (changes requested):
The year-selector should also take the year from the layer. Try the next:
* Browse to the analysis page
* In the selectors, chose a material (ex: Cereals)
* The layer will load and the year-selector is empty. It should be 2000 selected.-

Additionally, this branch has some conflicts :smile:

PR #279, opened 2022-03-01
added next auth secret env variable

davidsingal:
NextAuth.js is a library in the front end we use to manage the user session.
According to the [docs](https://next-auth.js.org/configuration/options#nextauth_secret), a new env variable `NEXTAUTH_SECRET` is required.


davidsingal:
@alexeh I was not sure if we also have to edit the file `kubernetes/modules/client/main.tf`

client/docs/installation.stories.mdx
```diff
@@ -10,6 +10,7 @@ Create a `.env.local` at the root of the app:
 
 ```
 NEXTAUTH_URL=http://localhost:3000
+NEXTAUTH_SECRET=nyancat
```
alexeh, line 13: lol nice

alexeh:
> @alexeh I was not sure if we also have to edit the file `kubernetes/modules/client/main.tf`

Whatever you need during image build, needs to be injected via GH. I understand what are you pointing out, actually we think those Kubernetes env vars are not required so once we have time to test it, we are gonna delete them and see if anything explodes

alexeh (approved):
LGTM, deploy it and see how it goes!

Thanks!

PR #286, opened 2022-03-03
Chore/landgrif 475 update queries for impact calculations

alexeh:


api/config/test.json
```diff
@@ -7,7 +7,7 @@
     "password": "",
     "database": "landgriffon_test",
     "synchronize": true,
-    "migrationsRun": false,
+    "migrationsRun": true,
```
tiagojsag, line 10: the default is "true", you can simply remove this line

api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -258,4 +264,133 @@ export class IndicatorRecordsService extends AppBaseService<
   async clearTable(): Promise<void> {
     await this.indicatorRecordRepository.delete({});
   }
+
+  /**
+   * @description Creates Indicator Records from all existing Sourcing Records in the DB
+   */
+
+  async createIndicatorRecordsForAllSourcingRecords(): Promise<any> {
+    const indicators: {
```
tiagojsag, line 273: create a type for this

api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -258,4 +264,133 @@ export class IndicatorRecordsService extends AppBaseService<
...
+      (indicator: { id: string; nameCode: string; h3DataId: string }) => {
+        indicatorMapper[indicator.nameCode] = indicator;
+      },
+    );
+    const indicatorRecords: IndicatorRecord[] = [];
+    calculatedData.forEach(
+      (calculatedIndicatorRecords: IndicatorRecordCalculatedValuesDto) => {
+        const indicatorRecordDeforestation: IndicatorRecord =
+          IndicatorRecord.merge(new IndicatorRecord(), {
+            value: calculatedIndicatorRecords[INDICATOR_TYPES.DEFORESTATION],
```
tiagojsag, line 295: there are multiple TS errors in this file, starting here.

api/src/modules/indicators/indicator.entity.ts
```diff
@@ -49,8 +49,9 @@ export class Indicator extends BaseEntity {
   @Column({ type: 'text', nullable: true, unique: true })
   shortName?: string;
 
+  // TODO: Make this non-nullable and update tests
   @Column({ type: 'text', nullable: true, unique: true })
```
tiagojsag, line 53: maybe remove nullable: true ?

tiagojsag (changes requested):

api/src/modules/indicators/indicator.entity.ts
alexeh: Yes it definitely should, but I'd wait to validate this as it will require to update a lot of tests, and since we rely 95% on e2e, lots not affected by these changes as well, is going to be a pain.

PR #293, opened 2022-03-09
Added profile page

AbdoKousara:
# Description

###  in this PR one page has been created 
   **`Profile` Page in `src/pages/profile` includes:**
- `_middleware` file
    - for authenticating the page
- `index` includes  two forms:   
   - one form for modifying the user data, full name & email, it send a request to `/users/me` endpoint
   - and one for changing user password,  it send a request to `/users/password` endpoint

### dependencies: 
 - `apiService` from 'services/api'  to send `patch` request to the API
  - `useCallback, useMutation` for handleing form submission and API response
      - `useMutation` has two cases onSuccess/onError when one of them happened will show alert to the user
- `yup, useForm` for  forms validation

Screenshot 2022-03-07 at 6 17 12 PM

### notes: 

[LANDGRIF-470](https://vizzuality.atlassian.net/browse/LANDGRIF-470)


client/src/containers/user-profile-forms/component.tsx
```diff
@@ -0,0 +1,107 @@
...
+      setIsLoading(false);
+
+      alert({
+        type: 'error',
+        title: 'An error has occured while saving. Please try gain later',
+      });
+
+      showAlert(true);
+    },
+  });
```
davidsingal, line 54: this should be moved to a new hook

client/src/containers/user-profile-forms/component.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const handleEditUserData = useCallback(
+    (data: UserProfilePayload) => {
+      reset();
+      setIsLoading(true);
```
davidsingal, line 59: why are you not using loading from mutation instead?

client/src/containers/user-profile-forms/component.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+      showAlert(true);
+    },
+
+    onError: () => {
+      setIsLoading(false);
+
+      alert({
+        type: 'error',
+        title: 'An error has occured while saving. Please try gain later',
```
davidsingal, line 49: The error should come from API message. This error could be only when message is undefined

client/src/containers/user-profile-forms/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+              </Button>
+            </div>
+          </form>
+        </div>
+      </div>
+    </section>
+  );
+};
+
+export default UserPasswordForm;
```
davidsingal, line 120: Same comments as file above

client/src/pages/profile/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <Transition
+            show={showAlert}
+            enter="transition-opacity ease-in-out duration-700 delay-1000"
+            enterFrom="opacity-0"
+            enterTo="opacity-100"
+            leave="transition-opacity ease-in-out duration-700 delay-1000"
+            leaveFrom="opacity-100"
+            leaveTo="opacity-0"
+          >
+            <Alerts items={alert} />
```
davidsingal, line 44: Alerts are more complicated to handle. It should be managed by redux and it should be added to the app. Maybe as Toaster notifications. But for now, let's leave it as it is.

davidsingal (changes requested):

PR #295, opened 2022-03-10
Multiple dropdowns selects child IDs instead of parents IDs

mluena:
**DESCRIPTION**

Multiple select behaviors changed. Now we storage just the ids of the children for multiple selector instead of using the parent one

**JIRA**
https://vizzuality.atlassian.net/browse/LANDGRIF-529?atlOrigin=eyJpIjoiYTg5ZjNiOTI4ZDA2NDBlMGE5MGNkOTIwYTAxMTc2NGIiLCJwIjoiaiJ9

### DIFF
--- a/client/src/containers/analysis-visualization/analysis-filters/more-filters/component.tsx
+++ b/client/src/containers/analysis-visualization/analysis-filters/more-filters/component.tsx
@@ -1,5 +1,8 @@
 import React, { Fragment, useCallback, useEffect, useState, useMemo } from 'react';
 import classNames from 'classnames';
+
+import { flatten } from 'lodash';
+
 import { Popover, Transition } from '@headlessui/react';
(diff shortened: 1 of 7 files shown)


client/src/containers/analysis-visualization/analysis-filters/more-filters/component.tsx
```diff
@@ -51,9 +54,16 @@ const MoreFilters: React.FC = () => {
   const handleChangeFilter = useCallback(
     // only save ids on store
     (key, values) => {
+      const childrenIds = values.map((v) => {
+        if (!!v?.children) {
+          return v.children;
+        }
+        return v;
+      });
```
davidsingal, line 62: why this is not inside the tree-select component? I think it could be

davidsingal (changes requested):

PR #299, opened 2022-03-14
Feature LANDGRIF-548 calculate impact for new interventions

alexeh:
This PR adds:

Service to calculate Indicator (Impact) Records for new Scenario Interventions.

Updates/Refactors some DTOs and implementation for the Scenario Intervention workflow to match some props and a little performance improvement

Better error handling on GeoCoding, to avoid caching location info that cannot be geocoded and raise and exception

**NOTE:** Tech debt on testing the service including the stored procedures, but you can review this while I am solving the debt

### DIFF
--- a/api/src/modules/import-data/sourcing-data/sourcing-data-import.service.ts
+++ b/api/src/modules/import-data/sourcing-data/sourcing-data-import.service.ts
@@ -109,7 +109,7 @@ export class SourcingDataImportService {
           dtoMatchedData.sourcingData,
         );
 
-      // TODO: TBD What to do when there is some location where we cannot determine it's admin-region: i.e coordinates
+      // TODO: TBD What to do when there is some location where we cannot determine its admin-region: i.e coordinates
       //       in the middle of the sea
       const geoCodedSourcingData: SourcingData[] =
         await this.geoCodingService.geoCodeLocations(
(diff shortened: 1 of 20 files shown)


api/src/modules/import-data/sourcing-data/sourcing-data-import.service.ts
```diff
@@ -108,6 +109,8 @@ export class SourcingDataImportService {
           dtoMatchedData.sourcingData,
         );
 
+      // TODO: TBD What to do when there is some location where we cannot determine it's admin-region: i.e coordinates
```
tiagojsag, line 112:
```suggestion
      // TODO: TBD What to do when there is some location where we cannot determine its admin-region: i.e coordinates
```

api/src/modules/import-data/sourcing-data/sourcing-data-import.service.ts
```diff
@@ -116,8 +119,21 @@ export class SourcingDataImportService {
       await this.sourcingLocationService.save(geoCodedSourcingData);
 
       this.logger.log('Generating Indicator Records...');
-      await this.indicatorRecordsService.createIndicatorRecordsForAllSourcingRecords();
-      this.logger.log('Indicator Records generated');
+
+      // TODO: Current approach calculates Impact for all Sourcing Records present in the DB
+      //       Getting H3 data for calculations is done within DB so we need to improve the error handling
+      //       TBD: What to when is no H3 for a Material
```
tiagojsag, line 125: I can no English derp

tiagojsag (commented):
TL;DR

api/src/modules/import-data/sourcing-data/sourcing-data-import.service.ts
alexeh: LOL I think I had a brain short-circuit

PR #309, opened 2022-03-21
Fix saving Geo Region h3Flat and h3FlatLength

yulia-bel:


api/test/integration/geo-coding/geo-coding.service.spec.ts
```diff
@@ -53,6 +57,79 @@ describe('GeoCoding Service (Integration Testing)', () => {
...
+        locationLongitude: 78.96288,
+        locationLatitude: 20.593684,
+      } as unknown as SourcingData;
+
+      const sourcingDataWithAddress = {
+        locationCountryInput: 'Asgard',
+        locationAddress: 'Valhalla, 3',
+      } as unknown as SourcingData;
+
+      expect.assertions(1);
```
alexeh, line 111: Is this test actually working OK? 
So the goal of this expect.assertions(number) is to actually expect a determined number of assertions (captain obvious here).

So in example: if you write a test with no ```expect(something).toSomething()``` The test will actually 'pass'.

This being said, if we expect a error to be thrown (that is inside a catch block scope), and there is no error thrown, it will never reach the expect statement and effectivelly will mean that 'there is no expect' in this test, therefore passing the test and giving us a false positive. ```expect.assertions(number``` is our guard here, if the test reach no expect, but if we state that we actually expect some expect (lol), the test will fail.

So I see 2 assertions here, but expecting just one. So either one of said expect is not being reached, or there is something wrong

Am I missing something?

api/src/modules/geo-regions/geo-region.repository.ts
```diff
@@ -30,6 +30,8 @@ export class GeoRegionRepository extends Repository<GeoRegion> {
       .createQueryBuilder()
       .select(`hashtext(concat($3::text, points.radius))`)
       .addSelect(`points.radius`)
+      .addSelect(`array(SELECT h3_polyfill(points.radius,6))`)
```
alexeh, line 33: I know that this full query is out of this PRs scope, but have a look how the (currently) 2 possible geo-regions are saved, query-wise. 

I think both of them should follow the same approach with the query, therefore improving readability and improving our overall QoL

Do you think it's worth refactoring one of them to both follow the same approach?

api/src/modules/geo-regions/geo-region.repository.ts
```diff
@@ -68,6 +70,9 @@ export class GeoRegionRepository extends Repository<GeoRegion> {
       .values({
         name: () => `hashtext(:arg1)`,
         theGeom: () => `ST_GeomFromText(:arg2, 4326)`,
+        h3Flat: () => `array(SELECT h3_geo_to_h3(ST_GeomFromText(:arg2), 6))`,
+        h3FlatLength: () =>
+          `cardinality(array(SELECT h3_geo_to_h3(ST_GeomFromText(:arg2), 6)))`,
         h3Compact: () =>
           `array( SELECT (h3_compact(array(SELECT h3_geo_to_h3(ST_GeomFromText(:arg2), 6)))))`,
```
alexeh, line 77: Again, I am aware this is out if this PR's scope, but one question. Does this query have a guard against a conflicting 'name' field, as the above query has?

alexeh:
Hi @yulia-bel nice work! I let some question, more than suggestions, let's try to clarify them and I'll be happy to proceed with this PR!

Thanks!

PR #312, opened 2022-03-22
Landgrif 519 distributed impact map proto

alexeh:
Test


elpamart:
This PR contains a first approach for the distribution impact optimized endpoint, using the new indicator record parameters:

- Scaler: total production for commodity x associated with the sourcing record
- Pointer: Id of commodity production h3TableName and h3TableColumn

The map distribution would be done by:

`Impcat(pixel level) = Production(pixel level) * Value / Scaler (Total production)`

NOTES:

- h3DataIndex type need to be updated.
- h3DataIndex currently pointing to the indicator table rather than the production h3table.
- Query needs to be optimized and updated.

elpamart:
Things that we need to update:

- General query should get the `h3DataId` from the `indicator_record ` rather from the `material_to_h3` entity
- The `h3DataId` type needs to be updated from text to uuid
- The `h3DataId`  needs to point to the production row no the indicator one as its doing right now.

PR #318, opened 2022-03-25
Update Sourcing Location queries to ignore interventions

yulia-bel:
Found 4 places where the 'no intervention' filter should be applied. The Sourcing Locations endpoint is questionable, probably @alexeh has different implementation for the comparison of the interventions, or mine will cause a conflict, since it excludes all the intervention Sourcing Locations from finadAllPaginated methos of the nestjs-base-service.

Also updated the import test with new Indicator tables names implementation. The fallback strategy tests are still skipped since new import implementation doesn't have it.


tiagojsag (approved):

alexeh (approved):
LGTM, thanks @yulia-bel !

I want to keep this PR in hold, until we sync with FE over some stuff, as I updated Prod with staging. Let's talk in Slack once you are here :)

PS: As you suggest, I might have to change some bits with the comparison table implementation, but nothing unexpected, thanks!

PR #322, opened 2022-03-28
Fixing missing data column

mluena:
Bug: Download button throwing an error after a successful request

Fix: data parse

**JIRA**
https://vizzuality.atlassian.net/browse/LANDGRIF-551

### DIFF
--- a/client/src/containers/admin/download-materials-data-button/component.tsx
+++ b/client/src/containers/admin/download-materials-data-button/component.tsx
@@ -29,7 +29,7 @@ const DownloadMaterialsDataButton: React.FC<DownloadMaterialsDataButtonProps> = 
     data: originalData,
   }: DownloaderTransformProps): DownloaderTransformProps => {
     const years = uniq(
-      flatten(originalData?.data?.map(({ purchases }) => purchases.map(({ year }) => year))).sort(),
+      flatten(originalData.map(({ purchases }) => purchases.map(({ year }) => year))).sort(),
     );
 
     const headers = [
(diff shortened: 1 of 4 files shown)


client/src/containers/admin/download-materials-data-button/component.tsx
```diff
@@ -29,25 +29,27 @@ const DownloadMaterialsDataButton: React.FC<DownloadMaterialsDataButtonProps> =
     data: originalData,
   }: DownloaderTransformProps): DownloaderTransformProps => {
     const years = uniq(
-      flatten(originalData.map(({ purchases }) => purchases.map(({ year }) => year))).sort(),
+      flatten(originalData?.data?.map(({ purchases }) => purchases.map(({ year }) => year))).sort(),
```
davidsingal, line 32: This is not wrong and it's fine. But could you also ensure this button is only available when the data is ready?

client/src/containers/downloader/component.tsx
```diff
@@ -44,16 +44,14 @@ const Downloader: React.FC<DownloaderProps> = ({
...
-            data && setData(data);
-            onSuccess();
-          })
-          .catch(onError);
+        getDataForExport().then(({ headers, data }) => {
+          headers && setHeaders(headers);
+          data && setData(data);
+          onSuccess();
+        });
+        // .catch(onError);
```
davidsingal, line 54: Why this error is commented? I think we need it

client/src/containers/downloader/types.d.ts
```diff
@@ -20,7 +24,7 @@ export type DownloaderProps = PropsWithChildren & {
   /** params to pass to the endpoint */
   params?: Record<string, string>;
   /** Custom headers for export */
-  headers?: DownloaderHeadersType;
+  headers?: { label: string; key: string }[];
```
davidsingal, line 27: `DownloaderHeadersType` was not correct?

davidsingal (changes requested):

PR #338, opened 2022-03-30
Update impact map query to use scaler + tests

yulia-bel:


api/src/migrations/1645259040554-ImpactStoredFunctions.ts
```diff
@@ -258,8 +258,53 @@ export class ImpactStoredFunctions1645259040554 implements MigrationInterface {
     END;
   $$
   LANGUAGE plpgsql;
-
   `);
+
+    await queryRunner.query(`
+    CREATE OR REPLACE FUNCTION get_h3_material_table_column(h3DataId uuid)
```
alexeh, line 264: Do we really need this stored function? Looking at the logic I think it would be better to have this on the codebase, but at the same time, since this is for retrieving a map, but getting just a unique table and column name by a ID should be fairly fast

api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -438,46 +435,59 @@ export class H3DataRepository extends Repository<H3Data> {
...
+
+    let sqlQuery: string =
+      query +
+      `, LATERAL(SELECT h3index, ir.value/ir.scaler * value as value FROM get_h3_data_over_georegion(sl."geoRegionId", mth."h3DataId")) h3data
+      WHERE sr.year=$1 
+      AND sl."scenarioInterventionId" IS NULL
+      AND sl."interventionType" IS NULL
+      AND ir."indicatorId"=$2`;
+
+    let numberOfParams: number = 2;
```
alexeh, line 459: what is this numberOfParams used for?

api/src/modules/h3-data/h3-data.repository.ts
```diff
@@ -438,46 +435,59 @@ export class H3DataRepository extends Repository<H3Data> {
...
 
     if (materialIds) {
-      query.andWhere('sl.material IN (:...materialIds)', { materialIds });
+      sqlQuery =
+        sqlQuery + ` AND sl."materialId" = ANY ($${numberOfParams + 1})`;
```
alexeh, line 463: wouldn't be easier to build up conditionally the query using the queryBuilder instead of manipulating a string? Feels a bit risky

api/src/migrations/1645259040554-ImpactStoredFunctions.ts
```diff
@@ -258,8 +258,53 @@ export class ImpactStoredFunctions1645259040554 implements MigrationInterface {
     END;
   $$
   LANGUAGE plpgsql;
-
   `);
+
```
alexeh, line 262: Stored functions added on this migration won't work, and I am really curious on how it worked for you (there is a reason but it is a weird one) 

Let's talk in slack about how migrations work :)

alexeh (changes requested):
Thanks @yulia-bel !

Apart from some questions I have regarding the implementation, I think there is a issue with the migration file. Let's talk in slack about how this works :)

alexeh (approved):

PR #350, opened 2022-04-05
Fix an issue where selectors are not showing current selection

JoaquinTrinanes:
Fixes select value not being shown when the value changed.
I tried adding some type safety to avoid this kind of bug in the future (`useForm<Step1Props>`), but I'm not exactly sure of the types of the other form elements and would need to check it out.

image

### DIFF
--- a/client/src/containers/interventions/new/step1/component.tsx
+++ b/client/src/containers/interventions/new/step1/component.tsx
@@ -168,7 +168,7 @@ const Step1: FC = () => {
             options={optionsBusinesses}
             placeholder="all businesses"
             theme="inline-primary"
-            onChange={({ value }) => handleDropdown('businessUnitsIds', value)}
+            onChange={(value) => handleDropdown('businessUnitsIds', value)}
             error={!!errors?.businessUnitsIds?.message}
           />
           <span className="text-gray-700 font-medium">from</span>
@@ -225,7 +225,7 @@ const Step1: FC = () => {
               current={watch('type')}
               options={optionsInterventionType}
               placeholder="Select"
-              onChange={(option) => handleDropdown('type', option)}
+              onChange={(value) => handleDropdown('type', value)}
             />
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


client/src/containers/interventions/new/step1/component.tsx
```diff
@@ -168,7 +168,7 @@ const Step1: FC = () => {
             options={optionsBusinesses}
             placeholder="all businesses"
             theme="inline-primary"
-            onChange={({ value }) => handleDropdown('businessUnitsIds', value)}
+            onChange={(value) => handleDropdown('businessUnitsIds', value)}
```
mluena, line 171: when you open the form for a new intervention the default option for this select should be "all business". Also, if you change the params for handleDropdown, you need to set the values differently in the function as we are saving just the business id (just the value of the option selected)

client/src/containers/interventions/new/step1/component.tsx
```diff
@@ -107,6 +109,8 @@ const Step1: FC = () => {
     dispatch(resetInterventionData());
   }, [dispatch]);
 
+  console.log(watch());
```
davidsingal, line 112: do not forget to remove console.log

davidsingal (changes requested):

davidsingal:
It's still not working, It's still failing on the selects of step 2.

client/src/containers/interventions/new/step1/component.tsx
JoaquinTrinanes: You're right, that's already taken care of. Next time I'll check that all data is stored correctly, promise!

client/src/containers/interventions/new/step1/component.tsx
JoaquinTrinanes: FYI ;) https://eslint.org/docs/rules/no-console

JoaquinTrinanes:
> It's still not working, It's still failing on the selects of step 2.

@davidsingal Should be fixed now!

davidsingal (approved):

PR #353, opened 2022-04-06
interventions templates

mluena:
Interventions should come from API and create a dynamic sentence depending on the intervention type. This PR adds different templates for the three intervention types


davidsingal:
Screenshot 2022-04-07 at 09 39 37

When you move the cursorr over an intervention, we don't have any UI feedback, a cursoir pointer and background change will be nice.

davidsingal (changes requested):
The code seems to be ok, but I found some bugs in the UI interaction

PR #354, opened 2022-04-06
Client/feature/landgrif 581 zoom controls

JoaquinTrinanes:
This PR adds zoom controls to the map, just above the legend
image

# Things to take into account
- When the legend opens the arrow is positioned wrong. Because David is working on redesigning it I left it as is to avoid conflicts as much as possible.
image

- As I explain in a comment, there's some weird behaviour with the map and the zoom level. DeckGL returns the `minZoom` as 0, but the map itself only zooms out until 1.28. Because of this, if the zoom is below than value the data is misaligned (as if the map were zoomed out). I hacked a fix hardcoding the true `minZoom` value, because I couldn't find a way to retrieve it otherwise.

### DIFF
--- a/client/src/components/map/controls/zoom/component.tsx
+++ b/client/src/components/map/controls/zoom/component.tsx
@@ -11,20 +11,27 @@ export interface ZoomControlProps {
   onZoomChange: (zoom: number) => void;
 }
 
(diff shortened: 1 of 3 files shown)


JoaquinTrinanes:
THis PR was already merged in #352

PR #355, opened 2022-04-07
new schema validation, business-units tree

mluena:
- New schema validation for intervention creation
- Selects in step 2 should appear empty when entering the form
- Business units in step 1 should allow multiple selections


client/src/components/steps/component.tsx
```diff
@@ -16,7 +16,9 @@ const Steps: React.FC<StepsProps> = ({ steps, current }: StepsProps) => {
   const handleTab = useCallback((step) => dispatch(setNewInterventionStep(step)), [dispatch]);
 
   const stepDescription = useCallback(
```
davidsingal, line 18: This should be a useMemo, or it could be also a component inside a constant.

davidsingal (changes requested):

davidsingal (approved):

PR #358, opened 2022-04-08
Add service to get Admin Regions descendants ids

yulia-bel:
The method should be used in the Impact table endpoint to process Admin Region ids (received as filters) and add the Ids of their descendants (if they exists) to the query.

To be added once the update of the response of Impact table endpoint is merged.

### DIFF
--- a/api/test/integration/admin-regions/admin-regions-get-descendants.spec.ts
+++ b/api/test/integration/admin-regions/admin-regions-get-descendants.spec.ts
@@ -1,15 +1,12 @@
 import { Test, TestingModule } from '@nestjs/testing';
-import { INestApplication } from '@nestjs/common';
 import { AppModule } from 'app.module';
 import { AdminRegion } from 'modules/admin-regions/admin-region.entity';
 import { AdminRegionsModule } from 'modules/admin-regions/admin-regions.module';
 import { AdminRegionRepository } from 'modules/admin-regions/admin-region.repository';
 import { createAdminRegion } from '../../entity-mocks';
-import { getApp } from '../../utils/getApp';
 import { AdminRegionsService } from 'modules/admin-regions/admin-regions.service';
 
 describe('AdminRegions - Get descendants by Admin Region Ids', () => {
-  let app: INestApplication;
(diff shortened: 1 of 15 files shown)


api/test/integration/admin-regions/admin-regions-get-descendants.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+      imports: [AppModule, AdminRegionsModule],
+    }).compile();
+
+    adminRegionRepository = moduleFixture.get<AdminRegionRepository>(
+      AdminRegionRepository,
+    );
+    adminRegionService = moduleFixture.get(AdminRegionsService);
+
+    app = getApp(moduleFixture);
+    await app.init();
```
alexeh, line 27: Improvement tip:

When we are not actually requiring the API to build a testing instance (like in this case, we are not sending requests and receiving responses over HTTP, there is no need to 
```
await app.init()
```

Even tho there is 'no harm' doing this, it's wasting resources and adding more runtime to the tests pipeline, since our pipeline is already quite heavy, we should avoid doing this as much as we can :)

api/test/integration/admin-regions/admin-regions-get-descendants.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    await adminRegionRepository.delete({});
+  });
+
+  afterEach(async () => {
+    await adminRegionRepository.delete({});
+  });
+
+  afterAll(async () => {
+    await app.close();
```
alexeh, line 36: And since there is no need for
```
await app.init()
```

There is no need for:
```
await app.close();
```

alexeh (approved):
Thanks @yulia-bel nicely done! I think this is a way better approach than mine for getting ancestry which is unnecesarily overcomplicated, I didn't know we could do this using TypeORM .findDescendants (and I guess the reverse using .findRoot)

Let's add refactoring my implementation to the tech debt

I left a very tiny improvements suggestion on the test side of things

Please remember to spread this functionality to the rest of the required entities, and connect your implementation to the whole workflow

PR #364, opened 2022-04-12
Changed tree-select strategy to show only parents

davidsingal:


davidsingal:
This PR depends on API progress to accept only parent IDs. @alexeh feel free to merge when you want

PR #381, opened 2022-04-25
fix impact table aggregation cases, when parent values where also inc…

asuramaru1:
Fix impact table aggregation cases, where parent values were also included when child ids were passed

### DIFF
--- a/api/src/modules/impact/impact.service.ts
+++ b/api/src/modules/impact/impact.service.ts
@@ -96,9 +96,19 @@ export class ImpactService {
           entitiesWithPagination.entities,
           fetchSpecification,
         );
-        impactTableDto.materialIds = this.getIdsFromTree(
+
+        /*
+         * Filter out ids to only include descendants of
+         * paginated root elements
+         */
+        const materialIds: string[] = this.getIdsFromTree(
           entitiesWithPagination.entities,
         );
+        impactTableDto.materialIds = impactTableDto.materialIds
+          ? impactTableDto.materialIds.filter((value: string) =>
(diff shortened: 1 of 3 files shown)


api/src/modules/impact/impact.service.ts
```diff
@@ -96,9 +96,19 @@ export class ImpactService {
...
+         * paginated root elements
+         */
+        const materialIds: string[] = this.getIdsFromTree(
           entitiesWithPagination.entities,
         );
+        impactTableDto.materialIds = impactTableDto.materialIds
+          ? impactTableDto.materialIds.filter((value: string) =>
+              materialIds.includes(value),
+            )
+          : materialIds;
```
KevSanchez, line 111: Considering this code seems to be a bit lower level to what the function getImpactTable does, it might be worth it refactoring this piece of code into the getIdsFromTree function (since it is a private function and only used three times for the same purpose)
It would receive the impactTableDto.materialsIds in this case, as an optional arugment, and do the filtering the same way inside the getIdsFromTree function

yulia-bel (approved):
Just a general comment for us, not to forget what we already discussed and have it as a Todo for better days: Probably it would make sense at some point to optimize the preconditions we have for the Impact tests (the fake data and relations in test database) - to be able to have more complex data set and cover more use cases, which now we sometimes miss, because of very limited fake entities for each test. (Probably, as Alex suggested, to have some sql script that will seed test database with something more close to the data we have in staging..)

KevSanchez (approved):

PR #382, opened 2022-04-26
Bugfix: LANDGRIF - 622 add only effective elements to intervention

alexeh:
**UPDATE:** 

After a chat with FR, we realised Supplierd Ids should not be required for creating an Intervention. Even tho this should be a separate task / PR, since it was depending on this implementation (if this goes forward) and for avoiding further changes, I've decided to include it here.

Let me know if this is smelly

**This PR fixes:**

Add only elements that are actually replaceable as part of the intervention: AdminRegions, Materials, etc. that are present in sourcing locations.

We were adding a full tree of elements given a Parent Id 

I created a standalone service to handle the creation of a newIntervention, which will take care of:

1. creating a new instance, adding required properties from the DTO
2. Get all descendants Id's for filtering
3. Add replaced elements
4. Add replacing elements

I created some tests accordingly

**IMPORTANT NOTE:**
Please if you are aware of some use case that is not covered with automated tests, please highlight it, we need to have this covered as much as possible, I'll be very happy to add more cases if something is missing

However, I am still not happy with the overall flow, I think we should discuss a way to separate concerns and make it more readable

Up to your thoughts @yulia-bel @KevSanchez

### DIFF
--- a/api/src/modules/scenario-interventions/dto/create.scenario-intervention.dto.ts
+++ b/api/src/modules/scenario-interventions/dto/create.scenario-intervention.dto.ts
@@ -170,7 +170,7 @@ export class CreateScenarioInterventionDto {
     enum: Object.values(LOCATION_TYPES),
     example: LOCATION_TYPES.POINT_OF_PRODUCTION,
   })
-  newLocationType?: LOCATION_TYPES;
+  newLocationType!: LOCATION_TYPES;
 
   @ValidateIf(
     (dto: CreateScenarioInterventionDto) =>
--- a/api/src/modules/scenario-interventions/dto/create.scenario-intervention.dto.ts
+++ b/api/src/modules/scenario-interventions/dto/create.scenario-intervention.dto.ts
@@ -103,8 +103,8 @@ export class CreateScenarioInterventionDto {
   businessUnitIds!: string[];
 
   @IsUUID(4, { each: true })
-  @IsNotEmpty()
-  @ApiProperty({
+  @IsOptional()
+  @ApiPropertyOptional({
     description:
       'Ids of Suppliers or Producers that will be affected by intervention',
     type: [String],
--- a/api/src/modules/scenario-interventions/dto/create.scenario-intervention.dto.ts
+++ b/api/src/modules/scenario-interventions/dto/create.scenario-intervention.dto.ts
@@ -45,7 +45,7 @@ export class CreateScenarioInterventionDto {
     enum: Object.values(SCENARIO_INTERVENTION_TYPE),
     example: SCENARIO_INTERVENTION_TYPE.NEW_MATERIAL,
   })
-  type!: string;
+  type!: SCENARIO_INTERVENTION_TYPE;
 
   @IsNumber()
   @IsNotEmpty()
--- a/api/src/modules/scenario-interventions/services/intervention-generator.service.ts
+++ b/api/src/modules/scenario-interventions/services/intervention-generator.service.ts
@@ -24,26 +24,6 @@ export class InterventionGeneratorService {
     protected readonly suppliersService: SuppliersService,
   ) {}
 
-  createInterventionInstance(
-    dto: CreateScenarioInterventionDto,
-  ): ScenarioIntervention {
-    const scenarioIntervention: ScenarioIntervention =
-      new ScenarioIntervention();
-    scenarioIntervention.title = dto.title;
-    scenarioIntervention.description = dto.description;
-    scenarioIntervention.scenarioId = dto.scenarioId;
-    scenarioIntervention.startYear = dto.startYear;
-    scenarioIntervention.percentage = dto.percentage;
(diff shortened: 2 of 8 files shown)


api/src/modules/scenario-interventions/dto/create.scenario-intervention.dto.ts
```diff
@@ -170,7 +170,7 @@ export class CreateScenarioInterventionDto {
     enum: Object.values(LOCATION_TYPES),
     example: LOCATION_TYPES.POINT_OF_PRODUCTION,
   })
-  newLocationType?: LOCATION_TYPES;
+  newLocationType!: LOCATION_TYPES;
```
yulia-bel, line 173: Sorry, I keep forgetting to change LOCATION_TYPES enum to have only the 4 location types we agreed on with Science ( UNKNOWN, AGGREGATION_POINT, POINT_OF_PRODUCTION, COUNTRY_OF_PRODUCTION), maybe we can do it in this PR to have this change sooner?

alexeh: Done! -> 5f1f110e90966c6427ac9603d8bec99b92bc92e9

api/src/modules/scenario-interventions/services/intervention-generator.service.ts
```diff
@@ -0,0 +1,157 @@
...
+import { SuppliersService } from 'modules/suppliers/suppliers.service';
+import { SourcingLocation } from 'modules/sourcing-locations/sourcing-location.entity';
+import { SourcingData } from 'modules/import-data/sourcing-data/dto-processor.service';
+
+/**
+ * @description: Utility wrapper to encapsulate some logics regarding new intervention generation
+ */
+
+@Injectable()
+export class InterventionGeneratorService {
```
KevSanchez, line 19: While I agree that that the original createScenarioIntervention function in ScenarioIntervention service was too big, and needed some functionality extracted, I think the approach could be slightly different than using another service/helper class, but not without deeper analysis for a bigger refactoring.
I would suggest using a Strategy pattern or similar as a basis for it, as it seems there's quite a bit of base code execution between each type.
In any case the current solution, makes the code more readable, and abstracts away some code, so it's fine for now :)

KevSanchez (changes requested):

api/src/modules/scenario-interventions/services/intervention-generator.service.ts
alexeh: I do totally agree that the whole workflow needs a deeper analysis for a proper refactoring. I'm all ears about what you might suggest for this. 
As you mentioned , you could push forward with this for the time being, but adding reviewing this as huge bullet point on our tech debt list

Let me know when you'd like to invest some time on this, as I think it will pay off

api/src/modules/scenario-interventions/services/intervention-generator.service.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  async addReplacedElementsToIntervention(
+    newIntervention: ScenarioIntervention,
+    cancelledSourcingLocations: SourcingLocation[],
+  ): Promise<ScenarioIntervention> {
+    const materialIds: string[] = [];
+    const adminRegionsIds: string[] = [];
+    const businessUnitIds: string[] = [];
+    const supplierIds: string[] = [];
+    for await (const sourcingLocation of cancelledSourcingLocations) {
```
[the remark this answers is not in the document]
alexeh: Good catch!!
Fixed -> [e058174](https://github.com/Vizzuality/landgriffon/pull/382/commits/e058174df54dffdae693a0f2fc02536b01bda6fe)

api/src/modules/scenario-interventions/services/intervention-generator.service.ts
```diff
@@ -0,0 +1,157 @@
...
+    scenarioIntervention.endYear = dto.endYear;
+    scenarioIntervention.type = dto.type;
+    scenarioIntervention.newIndicatorCoefficients =
+      dto.newIndicatorCoefficients as unknown as JSON;
+    scenarioIntervention.newLocationType = dto.newLocationType;
+    scenarioIntervention.newLocationCountryInput = dto.newLocationCountryInput;
+    scenarioIntervention.newLocationAddressInput = dto.newLocationAddressInput;
+
+    return scenarioIntervention;
+  }
```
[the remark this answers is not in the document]
alexeh: Thanks! Here the fix:
[6105583](https://github.com/Vizzuality/landgriffon/pull/382/commits/61055832ef947e0c0db18a8498a9552005f548a4)

KevSanchez (approved):

yulia-bel:
Regarding the tests, I think what might be useful at some point - to extends the preconditions to have more complex data set, probably the one we could use for analysis tests as well. Else, in my opinion, the possible use cases are covered

yulia-bel (approved):

PR #383, opened 2022-04-26
test(IndicatorRecord): Adds tests for Indicator Records

KevSanchez:
Adds integration tests for the createIndicatorRecordsBySourcingRecords method. Also refactors a bit of the code

### DIFF
--- a/api/src/modules/indicator-records/indicator-records.service.ts
+++ b/api/src/modules/indicator-records/indicator-records.service.ts
@@ -342,10 +342,9 @@ export class IndicatorRecordsService extends AppBaseService<
   }
 
   /**
-   * @description Creates Indicator-Records for each Sourcing-Record, by first retrieving Raw Indicator data from the DB, then applying
+   * @description Creates Indicator-Records for a single Sourcing-Record, by first retrieving Raw Indicator data from the DB, then applying
    * the methodology and persist new Indicator Records
    */
-
   async createIndicatorRecordsBySourcingRecords(
     sourcingData: {
       sourcingRecordId: string;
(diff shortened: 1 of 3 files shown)


api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -342,10 +342,9 @@ export class IndicatorRecordsService extends AppBaseService<
   }
 
   /**
-   * @description Creates Indicator-Records for each Sourcing-Record, by first retrieving Raw Indicator data from the DB, then applying
+   * @description Creates Indicator-Records for a single Sourcing-Record, by first retrieving Raw Indicator data from the DB, then applying
```
alexeh, line 345: This change is true for the time being, but it will change once we handle:

https://vizzuality.atlassian.net/browse/LANDGRIF-625

Worth changing it now or after said tasks is tackled?

api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -381,65 +380,51 @@ export class IndicatorRecordsService extends AppBaseService<
...
   }
 
+  /**
+   * Creates an IndicatorRecord object instance from the given input (type, calculated values, and h3data)
+   * @param indicatorType
+   * @param calculatedValues
+   * @param indicatorMapper
+   * @private
+   */
+  private static createIndicatorRecord(
```
alexeh, line 413: Why is this static? as suggested by the IDE?

api/test/integration/indicator-record/indicator-records.service.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+      await checkCreatedIndicatorRecord(
+        INDICATOR_TYPES.UNSUSTAINABLE_WATER_USE,
+        indicatorPreconditions.waterRisk,
+        sourcingData,
+        200,
+        null,
+      );
+    });
+
+    it("When creating indicator record with no provided coefficients, and there's no H3 data for the given material, it should throw an error it should create corresponding CANCELLED Sourcing Locations, with the appropiate Sourcing Records and Indicator Records", async () => {
```
alexeh, line 163: Very minor one: Can we keep using test() through all the project? Just for consistency

api/test/integration/indicator-record/indicator-records.service.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+      await indicatorRecordService.createIndicatorRecordsBySourcingRecords(
+        sourcingData,
+        providedCoefficients,
+      );
+
+      //ASSERT
+      const allIndicators = await indicatorRecordRepository.find();
+      expect(allIndicators.length).toEqual(4);
+
+      //Value is provided coeff * tonnage, and scaler is null, because it's based on production coefficient, and it's not provided
```
alexeh, line 132: Hm, not sure about this. I think since we've loaded both material and indicator h3 data as preconditions for this test suite, a production coefficient should be provided as well (and it will actually required for calculating a distributed impact map which is a WIP task) 

However, it will be possible that the production coefficient for a given indicator, for a given location, depending on the data can be 0, but I think it is weird that no Indicator Record has a value for production

Let's have a look into this together, please

api/test/integration/indicator-record/indicator-records.service.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+    );
+    expect(indicatorRecords[0].status).toEqual(INDICATOR_RECORD_STATUS.SUCCESS);
+    expect(indicatorRecords[0].value).toEqual(recordValue);
+    expect(indicatorRecords[0].scaler).toEqual(scalerValue);
+    expect(indicatorRecords[0].h3DataId).toEqual(h3Data.id);
+    //Inidicator Coefficientes are not checked because it's not used
+  }
+
+  async function deleteH3GridTables(): Promise<void> {
+    const tableList: any[] = await getManager().query(`select table_name
```
alexeh, line 281: I just loved this. It will tide us to use some fixed table names for preconditions, but since the createWorldForIndicator.... mock seed function is there and we are using it, I think it is safe for now leaving it

Just a note, should we extract these functionality to a external module? So we can use it in more suites

api/test/integration/indicator-record/indicator-records.service.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+                                                       from information_schema.tables
+                                                       where table_name like '%h3_grid_%'`);
+    if (tableList.length) {
+      const tableNameList = tableList.map((table: any) => table.table_name);
+      await getManager().query(`DROP TABLE ${tableNameList.join(', ')}`);
+    }
+  }
+
+  async function createPreconditions(): Promise<any> {
```
alexeh, line 290: Same for this, do you think it would be useful to have this available in a broader context, or do you think it's tied to this suite?

If it's tied to this context, maybe it will be a good idea to have it in a external file, what do you think?

api/test/integration/indicator-record/indicator-records.service.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+        INDICATOR_TYPES.UNSUSTAINABLE_WATER_USE,
+        indicatorPreconditions.waterRisk,
+        sourcingData,
+        12500,
+        10,
+      );
+    });
+  });
+
+  async function checkCreatedIndicatorRecord(
```
alexeh, line 259: Very nice approach, maybe we should extend this wherever possible

api/test/integration/indicator-record/indicator-records.service.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+
+      const sourcingData = {
+        sourcingRecordId: indicatorPreconditions.sourcingRecord2.id,
+        tonnage: indicatorPreconditions.sourcingRecord2.tonnage,
+        geoRegionId: indicatorPreconditions.sourcingLocation2.geoRegionId,
+        materialId: indicatorPreconditions.sourcingLocation2.materialId,
+      };
+
+      // The original function only expects for one MaterialTo3 data, the values don't matter
+      jest.spyOn(materialsToH3sService, 'findOne').mockResolvedValue({
```
alexeh, line 199: One note on the mocking side:

If you need to use the same mock for all the test cases, maybe it is better to use NestJs feature:

```
.overrideProvided(ClassToOverride).useClass(YourMockClass) 
```
or 
```
.overrideProvided(ClassToOverride).useValue(yourMockObject) 
```

Also, if you do need to mock different stuff, it is safer to use
```
jest.clearAllMocks()
```
In your afterEach() hook to be sure all mocks are clear within tests cases and avoid unexpected behaviours

alexeh (commented):
Thank you @KevSanchez I honestly liked your approach!

I left some comments that I would like to clarify before proceeding with this PR

Also, I liked how you simplified the calculation flow, do you think you can walk @yulia-bel and myself through your solution? I think both of us can learn something from your approach

Thanks!

api/test/integration/indicator-record/indicator-records.service.spec.ts
yulia-bel: +1, we could use the same approach to test Indicator Records created by Scenario Intervention

api/test/integration/indicator-record/indicator-records.service.spec.ts
KevSanchez: I see what you mean, but considering it's only mocked two times, I thought it maybe was worth it, when only mocking the result was needed.
It is true though that as it is right, is not properly the mocking result is not properly cleaned up. When using a mock function/value on a spied function, the mock will remain there until .restore() or jset-resotreAllMocks() is called. I should have used mockResolvedValue**Once**, that way once the mock has been executed once, the original behaviour will be used again. I'll change this for better, consistent behaviour.

api/test/integration/indicator-record/indicator-records.service.spec.ts
KevSanchez: It could be a nice to have to have some kind of generic function that checks one by one each property of a class, and depending on the input (if it's a complete set, or only a partial one) do expect checks as appropiate. Maybe jest already something like this. Although it might not be as trivial on more complex classes.

PR #384, opened 2022-04-26
Add drag functionality to the legend

JoaquinTrinanes:
Still need a drag handle design, for now a square should suffice :)

The impact layer is non-sortable and it will always be on top.


davidsingal:
Screenshot 2022-04-29 at 09 40 46

There is an issue with the alignment of the icon. Anyway, the design is ready: https://www.figma.com/file/hazwr4jct7j9Zeu99H3foJ/LG-Tool---Design-system?node-id=3291%3A2408

JoaquinTrinanes:
Thanks @davidsingal !

Should be fixed now, along with the new design ✨ 

image

davidsingal (approved):

PR #388, opened 2022-05-02
Editable year range

JoaquinTrinanes:
image


davidsingal:
Screenshot 2022-05-03 at 11 28 37

There is a bug, I've tried to create the year 2024. The list is not sorted well and it's not selected once is created.

Also, a question: what happens if the user tries to create a not valid year (a string)?

davidsingal (changes requested):
Now, when I create a new value, it doesn't select it by default

JoaquinTrinanes:
@davidsingal the component is functional, but styles are pending. All yours! :)

davidsingal (approved):

PR #390, opened 2022-05-05
Add location type filter to impact map and table

yulia-bel:


api/src/modules/impact/dto/get-impact-table.dto.ts
```diff
@@ -57,4 +58,10 @@ export class GetImpactTableDto {
   @IsUUID(4, { each: true })
   @Type(() => String)
   supplierIds?: string[];
+
+  @ApiPropertyOptional()
+  @IsOptional()
+  @IsEnum(LOCATION_TYPES, { each: true })
+  @Type(() => String)
+  locationTypes?: LOCATION_TYPES[];
```
alexeh, line 66: Shouldn't this be singular? Also can you please add relevant documentation to swagger to let the consumer know which are the available filters?

api/src/modules/impact/dto/paginated-entities.dto.ts
```diff
@@ -1,6 +1,12 @@
+import { LOCATION_TYPES } from 'modules/sourcing-locations/sourcing-location.entity';
 import { ImpactTableEntityType } from 'types/impact-table-entity.type';
 import { PaginationMeta } from 'utils/app-base.service';
 export class PaginatedEntitiesDto {
-  entities: ImpactTableEntityType[];
+  entities: ImpactTableEntityType[] | LocationTypeEntities[];
```
alexeh, line 5: Can't the new grouping option be an optional value of 
`ImpactTableEntityType` ? Instead of making the entities prop be optional. 
This is normally a bad practice and it should be avoided when possible

api/src/modules/impact/impact.service.ts
```diff
@@ -142,7 +146,7 @@ export class ImpactService {
           fetchSpecification,
         );
         impactTableDto.supplierIds = this.getIdsFromTree(
-          entitiesWithPagination.entities,
+          entitiesWithPagination.entities as ImpactTableEntityType[],
```
alexeh, line 149: Which is the error that you get to have to cast this, when you have made the entities prop be optionally one type or another?

alexeh (commented):
Thank you @yulia-bel !

LGTM, I have some (mostly) typing related questions that I'd want to clarify, I think we need to be more careful with the it

Feel free to hit me up through Slack to speed things up!

yulia-bel:
@alexeh thanks a lot for the review! I added the requested changes, tell me if it is better like this. Else we can look at the types details together, maybe come up with better solution..

api/src/modules/sourcing-records/sourcing-record.repository.ts
```diff
@@ -126,13 +136,30 @@ export class SourcingRecordRepository extends Repository<SourcingRecord> {
...
     }
     if (supplierIds) {
-      impactDataQueryBuilder
-        .andWhere('sourcingLocation.producerId IN (:...supplierIds)', {
-          supplierIds,
-        })
-        .orWhere('sourcingLocation.t1SupplierId IN (:...supplierIds)', {
-          supplierIds,
-        });
+      impactDataQueryBuilder;
```
KevSanchez, line 139: Just a small delete :)

api/test/e2e/impact/impact.spec.ts
```diff
@@ -1405,4 +1499,95 @@ describe('Impact Table and Charts test suite (e2e)', () => {
...
+          indicator,
+        });
+
+        await createSourcingRecord({
+          tonnage: startTonnage + 100 * index,
+          year,
+          indicatorRecords: [indicatorRecord],
+          sourcingLocation: sourcingLocation2,
+        });
+      }
```
KevSanchez, line 1571: Since the loop goes through the same years, maybe it's worth grouping both indicator and sourcing records into a single for loop, putting in some comments to the code more clear
`for await.....
    //Create Indicator and sourcing Record for country of production Sourcing Location
const startTonnage.....`

KevSanchez (changes requested):

PR #393, opened 2022-05-09
recreate scenario intervention when data which forces recalculation i…

asuramaru1:
…s passed on update

### DIFF
--- a/api/src/modules/scenario-interventions/scenario-interventions.service.ts
+++ b/api/src/modules/scenario-interventions/scenario-interventions.service.ts
@@ -37,6 +37,12 @@ export class ScenarioInterventionsService extends AppBaseService<
   UpdateScenarioInterventionDto,
   AppInfoDTO
 > {
+  private basicUpdateColumns: string[] = [
+    'title',
+    'description',
+    'updatedById',
+  ];
+
   constructor(
     @InjectRepository(ScenarioInterventionRepository)
     protected readonly scenarioInterventionRepository: ScenarioInterventionRepository,
@@ -141,6 +147,7 @@ export class ScenarioInterventionsService extends AppBaseService<
     const newScenarioIntervention: ScenarioIntervention =
       ScenarioInterventionsService.createInterventionInstance(
         dtoWithDescendants,
+        dto,
       );
 
     // Add replaced Entities to new Scenario Intervention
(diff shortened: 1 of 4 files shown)


api/src/modules/scenario-interventions/scenario-interventions.service.ts
```diff
@@ -443,7 +451,53 @@ export class ScenarioInterventionsService extends AppBaseService<
...
+    id: string,
+    dto: UpdateScenarioInterventionDto,
+  ): Promise<ScenarioIntervention> {
+    const currentScenarioIntervention: ScenarioIntervention =
+      await this.repository.findOneOrFail({ id });
+    const createScenarioDto: CreateScenarioInterventionDto = {
+      ...(currentScenarioIntervention.createDto as unknown as CreateScenarioInterventionDto),
+      ...dto,
+    };
+    const newScenario: ScenarioIntervention =
```
yulia-bel, line 494: Just a small naming comment - I think here it should be newScenarioIntervention, not newScenario - since Scenarios are different, more general, entities, that may have multiples ScenarioInterventions

yulia-bel:
Thanks @asuramaru1! I made a small comment about the naming, and I also have a question for everyone:
Why is it better to remove the existing Scenario Intervention and all its relations and create new ones instead of updating the existing ones? Is it possible that like this we will lose reference to the id of the original intervention which will be still needed somewhere? (I have no idea, just suggesting :) )

api/src/modules/scenario-interventions/scenario-interventions.service.ts
```diff
@@ -443,7 +451,53 @@ export class ScenarioInterventionsService extends AppBaseService<
     scenarioIntervention.newLocationType = dto.newLocationType;
     scenarioIntervention.newLocationCountryInput = dto.newLocationCountryInput;
     scenarioIntervention.newLocationAddressInput = dto.newLocationAddressInput;
+    scenarioIntervention.createDto = initialDto as unknown as JSON;
```
alexeh, line 454: Could this type casting be avoidable typing the property defined in the entity as CreatedInterventionDto?

asuramaru1: @alexeh at first i tried doing it like that, but that would cause circular dependency and bunch of errors.

KevSanchez (approved):

yulia-bel (approved):

PR #401, opened 2022-05-13
Fix tree selector filtering

JoaquinTrinanes:
This PR makes the tree select filter work internally instead of overriding the available options, preventing some bugs that marked the whole parent as selected when a single filtered child was selected.
image
Note how the Indonesia checkmark is only partially checked :)


client/src/components/tree-select/component.tsx
```diff
@@ -315,91 +305,110 @@ const TreeSelect: React.FC<TreeSelectProps> = ({
...
+                    onExpand={handleExpand}
+                    onSelect={handleSelect}
+                    onCheck={handleCheck}
+                    // filterTreeNode={(node) => {
+                    //   if (!searchTerm || searchTerm === '') {
+                    //     return true;
+                    //   }
+
+                    //   return Math.random() > 0.5;
+                    // }}
```
davidsingal, line 388: Remove this comments if it's not needed

davidsingal (approved):

PR #404, opened 2022-05-16
Add basic Cypress tests

JoaquinTrinanes:
This PR sets up basic Cypress test.

So far it tests:
- Can't login with incorrect password
- Can login with correct password
- Tree selectors can open, close and be clicked


client/cypress/config/cypress.local.json
```diff
@@ -0,0 +1,7 @@
+{
+  "baseUrl": "http://localhost:3000",
```
alexeh, line 2: Shouldn't this rely on config input?

PR #406, opened 2022-05-18
Client/feature/landgrif 615 layer tooltip info

JoaquinTrinanes:
image
The tooltip info is hardcoded in the front.


JoaquinTrinanes:
This is already implemented in #407

PR #414, opened 2022-05-24
Fix locationTypes naming and swagger

yulia-bel:
Location types as filters for Impact Map and Impact Table now in plural.
Swagger updated for Impact Map (available location types in description)


alexeh (approved):
Thanks a lot @yulia-bel for dropping everything to do is, it was needed. Really appreciated!

PR #415, opened 2022-05-24
storybook button styles

AbdoKousara:
Files have been created `Button.stories.tsx` in client/src/stories

new **Package** have been created for viewing buttons in different pseudo states
[storybook-addon-pseudo-states](https://storybook.js.org/addons/storybook-addon-pseudo-states)

***TASK IN JIRA***
[LANDGRIF-674](https://vizzuality.atlassian.net/browse/LANDGRIF-674)

### DIFF
--- a/client/src/components/button/component.tsx
+++ b/client/src/components/button/component.tsx
@@ -1,5 +1,5 @@
 import cx from 'classnames';
-import { forwardRef } from 'react';
+import { forwardRef, SVGProps } from 'react';
 import Loading from 'components/loading';
 import { PlusIcon } from '@heroicons/react/solid';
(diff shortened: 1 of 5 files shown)


client/src/components/button/component.tsx
```diff
@@ -1,15 +1,22 @@
 import cx from 'classnames';
 import { forwardRef } from 'react';
 import Loading from 'components/loading';
+import { PlusIcon } from '@heroicons/react/solid';
+import { MailIcon } from '@heroicons/react/outline';
```
davidsingal, line 5: This doesn't scale, it should become from outsie

davidsingal (changes requested):

PR #430, opened 2022-05-27
Basemap selector

JoaquinTrinanes:
This PR adds a toggle between default and satellite map.

One thing to note is that when switching to the default style the satellite layer is removed. When the same behaviour was implemented toggling it's visibility things started going weird: the map flickers, the tiles load again... I remove it for these reason.


davidsingal (changes requested):
Screenshot 2022-06-01 at 10 32 38

The main impact layer is behind the satellite layer. The basemaps should be always under the layers.

client/src/components/map/controls/basemap/component.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }, [onChange, value]);
+
+  const image = value === 'default' ? SatelliteImage : DefaultImage;
+
+  return (
+    <Tooltip
+      hoverTrigger
+      placement="right"
+      theme="dark"
+      content={<div className="bg-gray-900 text-white rounded-lg p-2">Change basemap!</div>}
```
davidsingal, line 31:
```suggestion
      content={<div className="bg-gray-900 text-white rounded-lg p-2">Change basemap</div>}
```

client/src/components/map/controls/basemap/component.tsx
```diff
@@ -0,0 +1,47 @@
...
+      onChange('satellite');
+    } else {
+      onChange('default');
+    }
+  }, [onChange, value]);
+
+  const image = value === 'default' ? SatelliteImage : DefaultImage;
+
+  return (
+    <Tooltip
```
davidsingal, line 27: Do we need a tooltip? Is it in the design??

client/src/components/tooltip/component.tsx
```diff
@@ -38,34 +47,59 @@ export const ToolTip: React.FC<TooltipProps> = ({
...
-      </Popover.Button>
-      <Popover.Panel
-        ref={floating}
-        className={className}
-        style={{
-          position: strategy,
-          top: y ?? '',
-          left: x ?? '',
+      <Popover.Button
+        onMouseEnter={() => {
```
davidsingal, line 62: consider `useCallback`

client/src/components/tooltip/component.tsx
```diff
@@ -38,34 +47,59 @@ export const ToolTip: React.FC<TooltipProps> = ({
...
+          setIsHovered(true);
+        }}
+        onMouseLeave={() => {
```
davidsingal, line 65: same, consider `useCallback`

davidsingal (changes requested):

davidsingal:
Also, it seems like the design changed a bit: https://www.figma.com/file/WxYMmJrUZAgV6f0Csmt8fK/LG-Tool---Work-in-progress?node-id=1724%3A40086

client/src/components/map/controls/basemap/component.tsx
JoaquinTrinanes: Removed it, also added the name of the basemap that will be changed to on swap

JoaquinTrinanes:
> Also, it seems like the design changed a bit: https://www.figma.com/file/WxYMmJrUZAgV6f0Csmt8fK/LG-Tool---Work-in-progress?node-id=1724%3A40086

That explains why I couldn't find the design 🥲

client/src/components/map/controls/basemap/component.tsx
JoaquinTrinanes: I just found the new design, I'll ask Andreia if we still need that tooltip

client/src/components/tooltip/component.tsx
JoaquinTrinanes: It won't probably matter, but I rather be consistent

client/src/components/map/controls/basemap/component.tsx
JoaquinTrinanes: image
Can confirm. I'll change the text accordingly 😁

davidsingal:
Screenshot 2022-06-06 at 11 15 52
Screenshot 2022-06-06 at 11 16 05

There is an issue with the tooltips (see screenshots).

davidsingal (changes requested):
I think the issue is not with z-index. It seems like opacity 1 is still opacity, and that's why with a different background it looks weird. Could you check it?

davidsingal (approved):

PR #435, opened 2022-05-30
Active layer number

JoaquinTrinanes:
This PR add a number specifying the number of current active layers. It also changes the style of the toggle layer button to an animated hamburger one.
image
image


davidsingal:
I really like the idea but the styles need to be reviewed. The number is not totally aligned at the center (at least in Chrome) and the button is a rectangle, the layers should be closer and maybe have a max number of layer because it doesn't scale very well taking on account the new design

davidsingal:
Screenshot 2022-05-31 at 09 57 33
My little contribution

PR #436, opened 2022-05-31
Feature/api/enchance geocoding

asuramaru1:
Enhance GeoCoding accuracy for more specific edge  cases

### DIFF
--- a/api/src/modules/admin-regions/admin-region.repository.ts
+++ b/api/src/modules/admin-regions/admin-region.repository.ts
@@ -7,7 +7,7 @@ import {
 import { AdminRegion } from 'modules/admin-regions/admin-region.entity';
 import { ExtendedTreeRepository } from 'utils/tree.repository';
 import { CreateAdminRegionDto } from 'modules/admin-regions/dto/create.admin-region.dto';
-import { Logger, NotFoundException } from '@nestjs/common';
+import { BadRequestException, Logger, NotFoundException } from '@nestjs/common';
 import { SourcingLocation } from 'modules/sourcing-locations/sourcing-location.entity';
 import { GetAdminRegionTreeWithOptionsDto } from 'modules/admin-regions/dto/get-admin-region-tree-with-options.dto';
 
@@ -117,7 +117,7 @@ export class AdminRegionRepository extends ExtendedTreeRepository<
       });
       const ancestor: AdminRegion = await this.findAncestorsTree(adminRegion);
       if (ancestor.name !== country)
-        throw new NotFoundException(
+        throw new BadRequestException(
           `coordinates ${coordinates.lng}, ${coordinates.lat} are not inside ${country}`,
         );
     }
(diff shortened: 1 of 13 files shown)


api/src/modules/admin-regions/admin-region.repository.ts
```diff
@@ -46,6 +49,17 @@ export class AdminRegionRepository extends ExtendedTreeRepository<
         `No Admin Region where Coordinates: LAT: ${searchParams.lat}, LONG: ${searchParams.lng} are could been found`,
       );
     }
+    if (country) {
```
alexeh, line 52: Maybe extract this logic to the service layer

api/src/modules/admin-regions/admin-region.repository.ts
```diff
@@ -91,10 +108,21 @@ export class AdminRegionRepository extends ExtendedTreeRepository<
...
-
-    return res.reduce(function (previous: any, current: any) {
+    const result: any = res.reduce(function (previous: any, current: any) {
       return previous.level > current.level ? previous : current;
     });
+    if (country) {
+      const adminRegion: AdminRegion = await this.findOneOrFail({
+        id: result.adminRegionId,
+      });
+      const ancestor: AdminRegion = await this.findAncestorsTree(adminRegion);
```
alexeh, line 118: Try if this works and avoid 2 roundtrips to DB
```suggestion
      const ancestor: AdminRegion = await this.findAncestorsTree({id: result.adminRegionId} as AdminRegion);
```

api/src/modules/geo-coding/strategies/point-of-production.geocoding.service.ts
```diff
@@ -25,12 +25,22 @@ export class PointOfProductionGeocodingStrategy extends BaseStrategy {
...
           },
         });
+      let adminRegionId: string;
 
-      const { adminRegionId } =
-        await this.adminRegionService.getClosestAdminRegionByCoordinates({
-          lng: sourcingData.locationLongitude,
-          lat: sourcingData.locationLatitude,
-        });
+      try {
```
alexeh, line 30: Check if this logic piece shared by different geocoding strategies can be externalised to the class extended by all

api/src/modules/import-data/sourcing-data/sourcing-data-import.service.ts
```diff
@@ -112,13 +114,21 @@ export class SourcingDataImportService {
...
           dtoMatchedData.sourcingData,
         );
-
       // TODO: TBD What to do when there is some location where we cannot determine its admin-region: i.e coordinates
       //       in the middle of the sea
       const geoCodedSourcingData: SourcingData[] =
         await this.geoCodingService.geoCodeLocations(
           sourcingDataWithOrganizationalEntities,
         );
+      const warnings: string[] = [];
```
alexeh, line 123: Double check if import should be stopped when a location is not accurately geocoded (multiple results)

alexeh: Jus checked that the import should continue adding the geolocation warning both to the Task keeping track of the import process and also to the Sourcing Location entity

PR #440, opened 2022-05-31
fixing bug, sidebar button expanding

AbdoKousara:
[LANDGRIF-678](https://vizzuality.atlassian.net/browse/LANDGRIF-678)


davidsingal:
Overall the PR is correct. I've reduced the animation time because should be the same as the sidebar animation. Additionally, I've updated the CHANGELOG. It will be a must from now.

PR #444, opened 2022-06-01
Improve-UX-login

AbdoKousara:
### General description

### Designs

### Testing instructions

## Checklist before merging

### DIFF
--- a/client/src/pages/auth/sign-in.tsx
+++ b/client/src/pages/auth/sign-in.tsx
@@ -6,6 +6,7 @@ import { signIn } from 'next-auth/react';
 import { useForm } from 'react-hook-form';
 import { yupResolver } from '@hookform/resolvers/yup';
 import * as yup from 'yup';
+import toast from 'react-hot-toast';
 
 import AuthenticationLayout from 'layouts/authentication';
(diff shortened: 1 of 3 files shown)


client/src/pages/auth/sign-in.tsx
```diff
@@ -32,9 +33,21 @@ const SignIn: React.FC = () => {
...
-      signIn('credentials', { ...data, callbackUrl: query.callbackUrl || '/analysis' });
+      signIn('credentials', {
+        ...data,
+        redirect: false,
+      }).then(({ error }) => {
+        if (error) {
+          setIsLoading(false);
+          setError('password', {
+            message: 'login failed. incorrect password',
+          });
```
davidsingal, line 44: Here you are assuming the error is because the password is wrong. We don't really know only with this piece of code. Anyway, the best, in this case, is to show a notification error as we have here: https://github.com/Vizzuality/landgriffon/blob/staging/client/src/containers/scenarios/component.tsx#L56

davidsingal (changes requested):

davidsingal:
Additionally, keep the changelog updated please

PR #447, opened 2022-06-02
Select styles

JoaquinTrinanes:
### General description

This PR overhauls the Select and TreeSelect styles so it matches the design system

### Designs

https://www.figma.com/file/hazwr4jct7j9Zeu99H3foJ/LG-Tool---Design-system?node-id=2703%3A1884

### Testing instructions

You can open up storybook, both selectors are added there!

## Checklist before merging


davidsingal:
As staging doesn't have data, we can see new cases to cover in the UI:

Screenshot 2022-06-03 at 09 24 22
here "No results" is missing

Screenshot 2022-06-03 at 09 24 46
here a placeholder

Screenshot 2022-06-03 at 09 25 03
and here we have "No results" but it should be in one line

client/tailwind.config.js
```diff
@@ -53,6 +53,7 @@ module.exports = {
         100: '#D1FAE5',
         50: '#EBF6F1',
       },
+      transparent: '#00000000',
```
davidsingal, line 56: what the hell is this?

JoaquinTrinanes: I added it at some time to avoid issues with borders. Going from no border to some border on focus resized the element. I'm not sure if I used it tho

davidsingal: let's try to remove it, at least it should have a length of 6

JoaquinTrinanes: Those extra numbers are the alpha value, the idea is for it to be invisible, not black or white

JoaquinTrinanes: I came upon some.uses of -transparent before adding that, and I remember that without it explicit in the config twin.macro doesn't detect it and throws an error

JoaquinTrinanes: I came upon some uses of -transparent before adding that, and I remember that without it explicit in the config twin.macro doesn't detect it and throws an error

davidsingal:
I think the PR is big enough to consider merging this branch. The missing issues or new bugs could be reported in new tasks and solved in other PRs.

PR #448, opened 2022-06-02
Chore: LANDGRIF-683 update model to store contextual layers

alexeh:
### General description

Create a new entity (ContextualLayers) to keep track of new magic h3 tables representing contextual layers
Related N to 1 to H3Data

Also added a new CHANGELOG file inside the /api folder. there is one at source level, should we keep the one outside, should we update the new one with the info the older one has?

Since the scope of this task is to just update the data-model to update the model so they can develop a script to populate it, I haven't add any test. 
Also I considered that this entity won't have a standard CRUD, besides probably getting category names and layer names
But since the needs here are still not clear, I think will be better to add specific test for specific tasks instead of adding
'useless' CRUD tests that a consumer will never use

### Designs

### Testing instructions

Run the API and check: 
1.a new table contextual_layers is created
2. Is related to H3Data
3. h3Data has a new column contextualLayerId

## Checklist before merging

### DIFF
--- a/api/src/modules/contextual-layers/contextual-layer.entity.ts
+++ b/api/src/modules/contextual-layers/contextual-layer.entity.ts
@@ -1,0 +1,60 @@
+import {
+  BaseEntity,
+  Column,
+  Entity,
+  ManyToOne,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+import { BaseServiceResource } from 'types/resource.interface';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { H3Data } from 'modules/h3-data/h3-data.entity';
+
+export enum CONTEXTUAL_LAYER_CATEGORY {
+  ENVIRONMENTAL_DATASETS = 'Environmental datasets',
+  BUSINESS_DATASETS = 'Business datasets',
+  FOOD_AND_AGRICULTURE = 'Food and agriculture',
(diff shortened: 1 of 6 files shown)


api/src/modules/contextual-layers/contextual-layer.entity.ts
```diff
@@ -0,0 +1,60 @@
...
+    plural: 'contextualLayers',
+  },
+  entitiesAllowedAsIncludes: [],
+  columnsAllowedAsFilter: ['name', 'metadata', 'category'],
+};
+
+@Entity()
+export class ContextualLayer extends BaseEntity {
+  @PrimaryGeneratedColumn('uuid')
+  @ApiProperty()
```
yulia-bel, line 32: Small question - since there is no controller and no endpoint for user to interact directly with that entity (so it won't be in swagger ui, if I understood correctly), do we need the openAPI tags like @ApiProperty() here?

alexeh: Good catch. We don't need a controller now since this task is to allow other people to actually use this new entity, but once we have clear requirements we will be using a controller for this module soon enough

yulia-bel:
Testing in local done - all good, just that thing we mentioned with h3dataCategory and h3data properties in the Entity

yulia-bel (approved):

PR #459, opened 2022-06-08
Bugfix: LANDGRIF-713 impact years filter by children material

alexeh:
### General description

Bugfix: When filtering available years for a Material in the Impact layer, a user can filter by a parent Material that is not actually present in sourcing locations. This is because in order to represent the entities in a tree and friendly format, we get the entities that are present in Sourcing Locations and build all the way up to the root.

There was a bug that we were not filtering by Materials that could be parent materials that are not present in Sourcing Locations, so they have no Sourcing Records, which represent the available years we have for generating Impact stuff, therefore sending back a empty array. We should filter also by ALL the children of the received Ids

So filtering by a Material, the API should return the available years for all the descendants that are present in Sourcing Locations, including itself (if it's present)

### Designs

### Testing instructions

Pretty much what is done in the automated tests. Create some Sourcing Locations and some materials, having a parent material that is not present in sourcing locations. 
Filter the API by said material, and you should receive the available years of it's children, which are actually present in sourcing locations and/so they have sourcing records

## Checklist before merging

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+##  2022-08-01
+### Added
+Fixed [LANDGRIF-713](https://vizzuality.atlassian.net/browse/LANDGRIF-713)
+Filter available years for a impact layer by the received Material Id's children
+
+
+
 ##  2022-06-01
 ### Added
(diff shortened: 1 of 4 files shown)


yulia-bel:
Tests in local done - all good 
Screenshot from 2022-06-08 17-56-39
Screenshot from 2022-06-08 17-55-40
Screenshot from 2022-06-08 17-54-55

KevSanchez (approved):

PR #462, opened 2022-06-09
Fix smart filters using applied values

JoaquinTrinanes:
### General description

This PR fixes the smart filters inside the "More filters" menu so they filter the data shown using their current selected values instead of the applied ones.

The `useQuery` options for this data fetching were also tweaked, adding them a 2 minute stale time. This way, we prevent refetching all data again every time the menu opens. This will probably happen in other places too, specially in dynamically  shown menus, but `useQuery` will automatically refetch everything stale (all queries by default) on window focus for example. We should look into this.

### Testing instructions

- Select Origin: Sri Lanka. The rest of the filters should be filtered. For example, the only available material should be silk.
- Remove the Origin filter. The filters should restore to their original state

## Checklist before merging

### DIFF
--- a/client/src/components/tree-select/component.tsx
+++ b/client/src/components/tree-select/component.tsx
@@ -403,25 +403,29 @@ const TreeSelect: React.FC<TreeSelectProps> = ({
               </div>
             )}
             {!loading && (
-              <Tree
-                autoExpandParent
-                checkStrictly={false}
-                checkable={multiple}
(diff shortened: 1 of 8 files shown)


alexeh (approved):
Just tested the deployed version, looks like works good! Thanks @JoaquinTrinanes !

davidsingal (approved):

PR #484, opened 2022-06-28
Fix bad requests

JoaquinTrinanes:
### General description

This PR fixes as bunch of bad requests that happened in some edge cases: intermediate states, `all` used as an ID...

### Testing instructions

- Open the analysis map and the network viewer
- Switch between the map, table and chart view
- Only calls to `https://events.mapbox.com/events/v2...` should fail

## Checklist before merging

### DIFF
--- a/client/CHANGELOG.md
+++ b/client/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 - An issue in the table component when the user navigates to the admin page
 - Prevent collapse button to break his position [LANDGRIF-729](https://vizzuality.atlassian.net/browse/LANDGRIF-729)
-
+- [LANDGRIF-716](https://vizzuality.atlassian.net/browse/LANDGRIF-716) fixed extra bad requests in analysis.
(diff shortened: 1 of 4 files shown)


client/.eslintrc.js
```diff
@@ -13,5 +13,6 @@ module.exports = {
         usePrettierrc: true,
       },
     ],
+    '@next/next/no-html-link-for-pages': ['error', 'client/src/pages'],
```
davidsingal, line 16: are you sure this is really needed?

davidsingal (approved):

client/.eslintrc.js
JoaquinTrinanes: Yeah, so [this rule](https://nextjs.org/docs/messages/no-html-link-for-pages) works correctly. You can see the warning in the ESLint output of the editor

davidsingal: Screenshot 2022-06-29 at 09 40 00
I'll remove it because I have the same warning with and without this rule. In case we need it, we can open a new PR only with this change.

PR #492, opened 2022-06-30
Update of creating new Scenario Intervention

yulia-bel:
### General description

Changes:

1. Scenario Intervention can affect only one existing  material (materialIds). However if the received material has descendant materials present in Sourcing Locations - those descendants will be added to filters when retrieving Sourcing Locations to be replaced. 
- 'materialIds' property of the dto remains an array for easier filter management, however it is validated that only 1 element is received 
2. businessUnitIds property for filters is not required. If no businessUnitIds are received, the filter won't be added to the query and sourcing locations of all business units (matching other filters)
3. 'adminRegionIds' property for filters is not required. If no businessUnitIds are received, the filter won't be added to the query and sourcing locations of all business units (matching other filters)
4. Location types format changed according to latest Location types endpoint: location types with dashes are received and transformed to database format
5. Update of test to cover changes

### Testing instructions

Expected endpoint behavior:

- Sending more than material to create intervention should return error message
- Intervention can be created without sending business unit ids and admin region ids
- Location types should be sent and validated in format with dashes

## Checklist before merging

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -51,8 +51,8 @@ Impact table - fix filters when retrieving 'group by' entities for table skeleto
 
 ## 2022-06-30
 
-### Fixed
+### Updated (change of requirements)
 
-Fixed
+Updated
 Creating Scenario Intervention can cover only 1 material, business units and admin regions filters are not obligatory
 
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -48,3 +48,11 @@ other smart filter endpoints.
 
 Fixed 
 Impact table - fix filters when retrieving 'group by' entities for table skeleton
+
+## 2022-06-30
+
+### Fixed
+
+Fixed
+Creating Scenario Intervention can cover only 1 material, business units and admin regions filters are not obligatory
+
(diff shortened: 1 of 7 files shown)


api/src/modules/scenario-interventions/dto/create.scenario-intervention.dto.ts
```diff
@@ -85,7 +91,10 @@ export class CreateScenarioInterventionDto {
   scenarioId!: string;
 
   @IsUUID(4, { each: true })
-  @IsNotEmpty()
+  @ArrayNotEmpty()
+  @ArrayMaxSize(1, {
```
alexeh, line 95: nice!

alexeh (approved):
Nice @yulia-bel! thanks for dropping everything and handling this, it was really needed!

I am happy with the code itself, waiting for @KevSanchez for review it. I would like to locally test the thing also, I'll do it ASAP, but since there are test also for the implementation, and we need this deployed, if @KevSanchez is happy with the thing, please feel free to move this forward!

api/CHANGELOG.md
```diff
@@ -43,3 +43,11 @@ other smart filter endpoints.
 
 Fixed 
 Impact table - fix filters when retrieving 'group by' entities for table skeleton
+
+## 2022-06-30
+
+### Fixed
```
alexeh, line 49: Very minimal, I believe this is a not actually a bug fix but a change of requirements. The previous implementation met the previous specs, so maybe it would be more accurate to reflect that as well, up to you

KevSanchez (approved):

PR #511, opened 2022-07-07
Fix impact table performance by updating getTrees methods 

yulia-bel:
### General description

Core entities (Material, Business Unit, Supplier, Admin Region) have methods to get those entities, that are present in Sourcing Locations in form of tree - get<Entity>TreeWithSourcingLocations(). 
Those options can accept filters (treeOptions) with ids of other entities and if such filter is received, the descendants of those filters are currently being found.

However, it is possible that there is no need to look for filters descendants as they were already searched and added to dto, before being sent to getTree() method, which leads to duplicated descendants search.

Example in Impact Service:

1. Dto with user filters is received (groupBy - material, filter - id of a large country)
2. All the filters pass through loadDescendantEntityIds() method. 
Now all the child regions of a large country are being added to initial dto to dto.originIds.
3. In order to find entities for the Impact table, trees with material names should be retrieved. Updated dto with countries children is passed to getEntityTree() method. 
4. Since case is 'group by material', all the large country's children will be passed as tree options to materialsService.getMaterialsTreeWithSourcingLocations(). Receiving them as treeOptions.originIds the method will once again search for the descendants (now not only of the original large country, but also for each and every of all its descendants previously found by loadDescendantEntityIds() method) - that is redundant and badly affecting performance..

on the other hand, when we use same get<Entity>TreeWithSourcingLocations() method for smart filters, we need them to be able to find descendant of filters.

Solution (Updated) - move the descendants search outside of get<Entity>TreeWithSourcingLocations() and make it responsibility of the function calling it.

### Testing instructions
Currently, when we request an Impact table, groupBy material, and add several large countries as filters (f.e. China, Australia, Canada) , data load takes very long time (due to excessive repeated descendants search).

With the fix - data is loaded many times faster.

## Checklist before merging

### DIFF
--- a/api/src/modules/materials/materials.service.ts
+++ b/api/src/modules/materials/materials.service.ts
@@ -170,9 +170,18 @@ export class MaterialsService extends AppBaseService<
 
   async getMaterialsTreeWithSourcingLocations(
     materialTreeOptions: GetMaterialTreeWithOptionsDto,
-    descendantsFound?: boolean,
   ): Promise<Material[]> {
-    if (!descendantsFound) {
+    const materialLineage: Material[] =
+      await this.materialRepository.getSourcingDataMaterialsWithAncestry(
+        materialTreeOptions,
+      );
+    return this.buildTree<Material>(materialLineage, null);
+  }
+
+  async getTrees(
+    materialTreeOptions: GetMaterialTreeWithOptionsDto,
+  ): Promise<Material[]> {
+    if (materialTreeOptions.withSourcingLocations) {
       if (materialTreeOptions.originIds) {
         materialTreeOptions.originIds =
           await this.adminRegionService.getAdminRegionDescendants(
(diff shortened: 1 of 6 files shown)


api/src/modules/materials/materials.service.ts
```diff
@@ -170,25 +170,29 @@ export class MaterialsService extends AppBaseService<
...
+          );
+      }
+      if (materialTreeOptions.supplierIds) {
+        materialTreeOptions.supplierIds =
+          await this.suppliersService.getSuppliersDescendants(
+            materialTreeOptions.supplierIds,
+          );
+      }
     }
+
```
KevSanchez, line 195: In order to avoid having to decide based on an external context of where the method is called (do the options have descendants ids or not), it would probably be better to move this responsibility to the caller, and, let the function do the processing on what it gets, and not make it more ambiguous based on an external context. This applies to the same variants of the function on each service (Materials, suppliers, etc)

KevSanchez (changes requested):

KevSanchez (approved):

PR #555, opened 2022-07-28
refactor(IndicatorRecord): Refactors a bit createAllIndicatorRecords

KevSanchez:
Just a little bit of cleanup, in preparation for when the createAllIndicatorRecordFromSourcingRecords is changed to be non dependent on stored functions, plus a new test

### General description

### Designs

### Testing instructions

## Checklist before merging

### DIFF
--- a/api/src/modules/indicator-records/indicator-records.service.ts
+++ b/api/src/modules/indicator-records/indicator-records.service.ts
@@ -194,7 +194,7 @@ export class IndicatorRecordsService extends AppBaseService<
           indicatorValues,
         };
 
-        return this.calculateIndicatorValuesV2(
+        return this.calculateIndicatorValues(
           sourcingRecordData.sourcingRecordId,
           sourcingRecordData.tonnage,
           sourcingRecordData.materialH3DataId,
@@ -269,7 +269,7 @@ export class IndicatorRecordsService extends AppBaseService<
           sourcingData.materialId,
           sourcingData.year,
         );
-      calculatedIndicatorRecordValues = this.calculateIndicatorValuesV2(
+      calculatedIndicatorRecordValues = this.calculateIndicatorValues(
         sourcingData.sourcingRecordId,
         sourcingData.tonnage,
         materialH3Data.id,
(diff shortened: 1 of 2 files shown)


api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -101,7 +101,6 @@ export class IndicatorRecordsService extends AppBaseService<
           `Cannot calculate impact for sourcing record - missing ${type} h3 data for material with ID "${sourcingRecord.materialId}" and year "${sourcingRecord.year}". Ignoring souring record`,
```
alexeh, line 101:
```suggestion
          `Cannot calculate impact for sourcing record - missing ${type} h3 data for material with ID "${sourcingRecord.materialId}" and year "${sourcingRecord.year}". Ignoring sourcing record`,
```

api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -156,69 +154,70 @@ export class IndicatorRecordsService extends AppBaseService<
    */
   // TODO still not adapted to modular indicators, because the performance gets hit drastically on the source import. Pending to be worked on
   async createIndicatorRecordsForAllSourcingRecords(): Promise<void> {
+    const indicatorMap: Record<string, any> = await this.getIndicatorMap();
+    // Using a hardcoded list of indicator types because this function is still based on
+    // a calculation query with hardcoded indicators,
+    const indicators: INDICATOR_TYPES[] = [
```
alexeh, line 160: Maybe something like 
```
Object.keys(INDICATOR_TYPES)
```
?
However, shouldn't we perform calculations for indicators that are only available in the DB? Or is this handled at some other point?

api/src/modules/indicator-records/indicator-records.service.ts
```diff
@@ -156,69 +154,70 @@ export class IndicatorRecordsService extends AppBaseService<
...
+          sourcingRecordData.rawWater,
+        );
+
+        const indicatorComputedRawDataDto: IndicatorComputedRawDataDto = {
+          harvestedArea: sourcingRecordData.harvestedArea,
+          production: sourcingRecordData.production,
+          indicatorValues,
+        };
+
+        return this.calculateIndicatorValuesV2(
```
alexeh, line 197: Can't we effectively refactor this to 
```suggestion
        return this.calculateIndicatorValues(
```
And remove the 'old' approach?

alexeh (commented):
Thanks a lot @KevSanchez for this!

I don't have objections but questions about the implementation. Also I am very interested on learning how the caching works regarding the intervention calculus.

Maybe we can set some meeting to clarify this points? Let me know

PR #560, opened 2022-08-01
New table design

JoaquinTrinanes:
### General description

This PR implements a total redesign of the tables using `@tanstack/react-table` in lieu of `ka-table`.
### Designs

[Analysis table](https://www.figma.com/file/Bdxq4bJD6eXFG1BuSjw6gG/Landgriffon---Redesign?node-id=478%3A19064)
[Admin table](https://www.figma.com/file/Bdxq4bJD6eXFG1BuSjw6gG/Landgriffon---Redesign?node-id=1044%3A74858)

## Checklist before merging


davidsingal:
It looks great, but some tweaks could be in another task:

* When the user interacts with the pagination it shouldn't change the height of the table
* for any new fetch there should be a spinner
* it's missing a space at the top of the table

Why do we only have sort for the first column?? could it be for more columns?

JoaquinTrinanes:
> It looks great, but some tweaks could be in another task:
> 
> * When the user interacts with the pagination it shouldn't change the height of the table
> * for any new fetch there should be a spinner

Yep, I pass the loading state but don't use it D:
Question: the old data should show while loading right? But I guess this should be done from the outside with the react-query helpers. This is in relation to the data disappearing when the page changes and such

> * it's missing space at the top of the table

That's true, the old one had it but I created this one without it because it's supposed to fit anywhere. I'll add it from the outside
> * do not cut the values, in some values I can see `9...`

So the column size width should ALWAYS match the content, no matter how long?
> * bug: the last page buttton in the pagination doesn't work and select the previous one

Oh no! Looking into it :)
> 
> Why do we only have sort for the first column?? could it be for more columns?

It could be for any column! I made it opt-in but we can change it. It's enabled passign the column the `enableSorting` prop

davidsingal (approved):

PR #563, opened 2022-08-02
Bugfix: LANDGRIF-780 solve gadm boundary issues

alexeh:
### General description

This PR adds a fix to solve GADM accuracy issues.

To validate determine and validate an admin region given address or coordinates, it creates a 1KM buffer given this geometry to intersect with all admin regions.

This 1KM buffer should be enough to solve this lack of accuracy. If the country provided by the user is within the intersecting countries of this buffer, we consider that the location is accurate

It also brings back this validation for interventions

Plus adds a test for when the geocoder returns multiple results and we save the most accurate interpretation

 

### Designs

### Testing instructions

There are 2 new files in the repo that you can use for testing. 

1. The one named 'multiple results' should create 4 sourcing locations, with 2 of them having a location warning value in sourcingLocations.locationWarning, as these sourcing location geolocation data returns multiple results (A region and a city withing that region could be named equaly for example)

2. The 'address outside country' file has a location with a spanish address, but Japan as a country. This should raise and exception, letting the user know this info. Remember you can't have this info in a sync manner if you run this by Postman.
You can check the logs, and you can also check the error in the Task tracking this async import

## Checklist before merging


KevSanchez (approved):

yulia-bel:
Testing in local done - all as expected 
Screenshot from 2022-08-03 16-31-13
Screenshot from 2022-08-03 16-29-58
Screenshot from 2022-08-03 16-28-57

yulia-bel (approved):

PR #568, opened 2022-08-05
[Chore] Upgrade to Next.js 12 and React 18

JoaquinTrinanes:
### General description

The main change resides in the middleware. Instead of using a `_middleware.ts`
 file perprotected page, there's now a main, global `middleware.ts`. There, the pages that need authentication can be set with:
```ts
export const config = {
  matcher: ['/admin/:path*', '/analysis/:path*'],
};
```

I tried to make the middleware secure by default, but ran into a bunch of issues. First of all, ALL requests go through it. That includes API calls, assets (those in the `public` folder like images)&hellip; It's a mess to safelist everything, so I kept the default behaviour of just protecting whatever matches the matchers.

Once we've got user roles, we can easily protect routes based on the role with some simple changes:
```ts
// src/middleware.ts
import { withAuth } from 'next-auth/middleware';

export default withAuth({
  callbacks: {
    authorized: ({ req, token }) => {
      if (!token) return false;

      if (req.nextUrl.pathname.startsWith('/admin')) {
        return token.isAdmin;
      }

      return true;
    },
  },
});

export const config = {
  matcher: ['/admin/:path*', '/analysis/:path*'],
};
```
## Checklist before merging


client/package.json
```diff
@@ -63,26 +63,26 @@
...
+    "react-dom": "18.2.0",
     "react-dropzone": "11.5.1",
     "react-hook-form": "7.27.1",
     "react-hot-toast": "2.2.0",
     "react-map-gl": "6.1.15",
-    "react-query": "3.34.16",
+    "react-query": "^3.39.2",
     "react-range": "1.8.12",
-    "react-redux": "7.2.4",
+    "react-redux": "^8.0.2",
```
andresgnlez, line 85: @JoaquinTrinanes fix the versions of `react-query` and `react-redux`, please

andresgnlez:
@JoaquinTrinanes `http://localhost:3000/analysis?scenarioId=%22actual-data%22&map=%7B%22longitude%22%3A0%2C%22latitude%22%3A0%2C%22zoom%22%3A2%7D` triggers this error:

image

andresgnlez (approved):

PR #578, opened 2022-08-23
Separate analysis pages

JoaquinTrinanes:
### General description

This PR separates the map, table and chart views in different pages with their own URL.

Key takeaways:
- Added `next-seo` to control the title and title template of the pages
- Used Next.js' [official way of adding layout support](https://nextjs.org/docs/basic-features/layouts). I really want the RFC to come though, this is horrible and I'm in pain. The method they recommend makes layouts not have any context, and that's a big issue because we use query providers, redux providers... I tweaked it so layouts are a normal react component that accept children and it appears to work fine :)
- Still store the current visualization mode in the Redux store, as there's a bunch of places that depend on it.
- `/analysis` isn't a valid URL anymore, so I added a redirect to `/analysis/map`

## Checklist before merging


andresgnlez:
Well done!

andresgnlez (approved):

PR #579, opened 2022-08-24
adds test for intervention creation (switch to material case)

andresgnlez:
### General description

Adds test for intervention creation (only Switch to material flow is covered so far).

To avoid fulfilling the API with test mocks, I am intercepting the requests to the API until we have a proper environment for that.

### Designs

### Testing instructions

`yarn 2e2` or `yarn e2e:headless`.

## Checklist before merging


alexeh (approved):
Hey @andresgnlez thanks a lot for this!

Just a couple of notes:

In [client/cypress/e2e/intervention-creation.cy.ts]
(https://github.com/Vizzuality/landgriffon/pull/579/files#diff-1b7fd09037a588f4bbc19d0334de11fdca87adde309134620ce9bce23179fc0a)
You are at some point asserting that you receive a 201 status code from a POST request and also checking some id, but then you are intercepting this request.
So I believe you are effectively testing that Cypress does what is meant to do when intercepting request, I honestly think we should focus 'outside-in', just our part of the code.

I.e: Asserting that after a successful request, a happy toast is shown (which you do just 2 lines below), and then we could also assert that after a failed response, an error toast with a fixed message (depending on the fixture you send back) is shown. I see much more value in that, and then we can build our testing suite following this philosophy, but we should really focus on not testing third libraries/frameworks, unless mandatory.

Also, this wouldn't be helpful in case we migrate to a real environment, where you can't know the ID of the resource as you are the one creating in for the assertion

Just my two cents, thanks a lot for contributing to this!

andresgnlez:
> So I believe you are effectively testing that Cypress does what is meant to do when intercepting request, I honestly think we should focus 'outside-in', just our part of the code.

I.e: Asserting that after a successful request, a happy toast is shown (which you do just 2 lines below), and then we could also assert that after a failed response, an error toast with a fixed message (depending on the fixture you send back) is shown. I see much more value in that, and then we can build our testing suite following this philosophy, but we should really focus on not testing third libraries/frameworks, unless mandatory.

I do agree, I'll focus on how the application behaves after a successful/failed request from the API. 👍🏻 

> Also, this wouldn't be helpful in case we migrate to a real environment, where you can't know the ID of the resource as you are the one creating in for the assertion

Yes, I am aware in the pipeline workflow, checking against the running API and not intercepting requests would be the best scenario (and we also ensure the API behaves as expected), but if we do this, there's no way we can run the test locally without fulfilling the real API (staging one, presumably) with dummy data coming from tests execution.

Not sure how to deal with this, to be honest. We need to be able to run the tests locally in order to write them, but I understand right now the tests aren't real e2e ones as we mock any external communication.

PR #607, opened 2022-09-06
Persistent smart filters

JoaquinTrinanes:
### General description

This PR filters the analysis filters being reset between pages.
This happened because of a combination of multiple factors:
- The analysis views are now in separate pages. This isn't the problem, as the layout is persistent and everything's in the store. It's just the detonator of the other issues
- The store was memoized in a way that it was reset every time the page changed
- Some filters overrode the previous one even if it had a value

All those issues are allegedly solved now :)

## Checklist before merging


client/CHANGELOG.md
```diff
@@ -7,11 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
...
+### Added
+
 ### Changed
 
 - Same input for city, address and coordinates for intervention form [LANDGRIF-821](https://vizzuality.atlassian.net/browse/LANDGRIF-821)
 - Redesign the Legend [LANDGRIF_727](https://vizzuality.atlassian.net/browse/LANDGRIF-727)
 
+### Fixed
+
+- Smart filters not being persisted between pages
```
davidsingal, line 19: remember to include the task link

client/src/components/table/component.tsx
```diff
@@ -12,7 +12,7 @@ import Cell, { HeaderCell } from './cell';
 import type { ColumnDefinition } from './column';
 import TableRow, { TableHeaderRow } from './row';
 
-export interface TableProps<T = unknown>
```
davidsingal, line 15: ensure @andresgnlez's PR doesn't have a conflict here because he was changing this files. Anyway this change couldn't be part of the feature

client/src/pages/_app.tsx
```diff
@@ -27,7 +25,8 @@ type AppPropsWithLayout = AppProps & {
 };
 
 const MyApp: NextPage<AppPropsWithLayout> = ({ Component, pageProps }) => {
-  const store = useMemo(() => initStore(pageProps.query), [pageProps.query]);
+  const [store] = useState(() => initStore(pageProps.query));
+  const [queryClient] = useState(() => new QueryClient());
```
davidsingal, line 29: this change should be included in the changelog, add a comment explaining why it was moving inside

davidsingal (changes requested):
It's ok but you are still combining different features in the same PR. In that case, at least try to include them in the changelog

davidsingal:
a part of my comments, the issue was fixed and works fine. Do you think could you add a test??

PR #636, opened 2022-09-20
pass h3 version to entrypoint

alexeh:
### General description

Pass H3 version to entrypoint.sh running a probably redundant pgxn command. Please check if both are actually required

### Designs

### Testing instructions

Run locally the image. With this code this CI run should also pass. Check after merging that CI is running ok

## Checklist before merging

### DIFF
--- a/database/Dockerfile
+++ b/database/Dockerfile
@@ -14,4 +14,4 @@ RUN pgxn install "h3==${PG_H3_VERSION}"
 
 COPY entrypoint.sh /usr/local/bin/
 
-ENTRYPOINT ["entrypoint.sh"]
+ENTRYPOINT entrypoint.sh "${PG_H3_VERSION}"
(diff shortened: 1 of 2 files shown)


BielStela (commented):
maybe also add the `--yes` options so that in case it breaks again we only see the pgxn error message and not the full python exception trace?

BielStela (approved):
:1st_place_medal:

alexeh:
> maybe also add the `--yes` options so that in case it breaks again we only see the pgxn error message and not the full python exception trace?

I agree, I forgot the pgxn yes issue in the middle of the chaos. 

Added in [fe444e5](https://github.com/Vizzuality/landgriffon/pull/636/commits/fe444e5da2b20489da44a0978bfed9641e0d8002)

tiagojsag:
use https://github.com/Vizzuality/landgriffon/pull/637 instead

PR #662, opened 2022-10-05
Fix initial store state

JoaquinTrinanes:
### General description

This PR deletes `ACTUAL_DATA` in favour of a null scenario id.

#### Actual data

Rationale:
- The current scenario ID had a default value of `null`
- When showing the scenario selector, a effect automatically set it to `ACTUAL_DATA` if not set. Hence, there's no distinction between a null id and a `actual-data` id
- The difference in types caused a bunch of edge cases. For example, when calling some endpoints checks had to be performed to exclude the scenario id if it was `actual-data`.
  ```ts
  const vsActual = !baseScenarioId || baseScenarioId === ACTUAL_DATA.id;
  ```
  ```ts
  const params = {
    // snip
    ...(currentScenario !== ACTUAL_DATA.id ? { scenarioId: currentScenario } : {}),
  };
  ```
  
  This cases are now simplified.

The only difficulty I found implementing this was knowing when to enable comparison mode. Before, it was done trivially checking if both `baseScenarioId` and `scenarioToCompare` were set. Now, it is done in the change handler of the `scenarioToCompare` selector.

#### The bug itself

The bug was caused by a surprisingly large amount of factors (every one of this caused the bug on it's own):
- The store initialisation didn't use the current values, only the default ones. So whatever value was present in the past was overriden
- This wasn't noticed because the store didn't refresh at all after the first load. Hence, even when navigating around and clicking links with query params, those new params weren't stored.

## Checklist before merging

### DIFF
--- a/client/CHANGELOG.md
+++ b/client/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
+- `ACTUAL_DATA` was removed and a null scenario id is used in it's place
 - Send `scenarioId` param alongside analysis filters [LANDGRIF-891](https://vizzuality.atlassian.net/browse/LANDGRIF-891)
 - Same input for city, address and coordinates for intervention form [LANDGRIF-821](https://vizzuality.atlassian.net/browse/LANDGRIF-821)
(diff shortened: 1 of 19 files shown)


davidsingal:
Everything seems to be working well. Only one little thing: when you select a scenario in the table view and change it to a chart, the selection doesn't appear in the URL. This selection should be shared among those pages (chart, table and map)

PR #695, opened 2022-10-17
Feature: LANDGRIF-959-Add dynamic indicator coefficient validation (part 1)

alexeh:
### General description

This PR adds the first piece of the task:
[Calculate Impact only for active indicators](https://vizzuality.atlassian.net/jira/software/c/projects/LANDGRIF/boards/14?selectedIssue=LANDGRIF-959)

Adding a dynamic validator at controller level to validate that received coefficients for indicators in the intervention creation flow matches the active indicators in the database

Adds a new test file (intervention controller validation) aiming to separate different tests so it can be more readable and also easier to mock when we can actually avoid building heavy preconditions

### Designs

### Testing instructions

To manually test this feature you'll need to set to true USE_NEW_METHODOLOGY env var and randomly set some indicators to true and others to false.

the scenarioIntervention.newIndicatorCoefficient should be validated agains the nameCodes of the active indicators.

See added tests for more ingo

## Checklist before merging


api/src/modules/indicator-records/services/impact-calculator.service.ts
```diff
@@ -124,13 +125,12 @@ export class ImpactCalculator {
...
-    const indicatorNameCodes: INDICATOR_TYPES_NEW[] =
-      Object.values(INDICATOR_TYPES_NEW);
+
     const indicatorsToCalculateImpactFor: Indicator[] = await getManager()
       .createQueryBuilder()
       .select()
       .from(Indicator, 'i')
-      .where('i.nameCode IN (:...nameCodes)', indicatorNameCodes)
+      .where('i.status = :status', { status: INDICATOR_STATUS.ACTIVE })
       .getMany();
```
KevSanchez, line 134: I would probably refactor this into a getActiveIndicators or something like that in the Indicator service/repository
The function in the service could construct a base options object as { status: active} and then merge the incoming options, and use those for a normal indicatorRepository.find()

KevSanchez (changes requested):

api/src/modules/indicator-records/services/impact-calculator.service.ts
alexeh: Totally agree, I left this at is was to have any dependencies and to be as much isolated as possible while we validate it, but I guess we should refactor it from here

PR #718, opened 2022-10-25
Fix/landgrif 977 past years projection

yulia-bel:
### General description

Changes the logic for isProjected properties in impact responses, depending on the year:

- if there is data for the year - isProjected: false
- if there is no data and the year is in the past - value: 0, isProjected : false
- if it is the year after the last year with available dat - value: projected in accordance with growth rate, isProjected: true

Test updated accordingly

### Designs

### Testing instructions

## Checklist before merging


api/src/modules/impact/comparison/actual-vs-scenario.service.ts
```diff
@@ -297,13 +302,16 @@ export class ActualVsScenarioImpactService {
                     rowValuesIndex - 1
                   ].scenarioValue || 0
                 : 0;
+
+            const isProjected: boolean =
+              year > lastAvailableDataYear ? true : false;
```
yulia-bel, line 307: const isProjected: boolean = year > lastAvailableYear

api/src/modules/impact/comparison/actual-vs-scenario.service.ts
```diff
@@ -376,6 +384,7 @@ export class ActualVsScenarioImpactService {
       this.getTotalPurchasedVolumeByYear(
         rangeOfYears,
         dataForActualVsScenarioImpactTable,
+        lastAvailableDataYear,
```
yulia-bel, line 387: lastYearWithData

api/src/modules/impact/comparison/actual-vs-scenario.service.ts
```diff
@@ -415,10 +425,12 @@ export class ActualVsScenarioImpactService {
             : 0;
         const tonnesToProject: number =
           dataForImpactTable.length > 0 ? previousYearTonnage : 0;
+        const isProjected: boolean =
+          year > lastAvailableDataYear ? true : false;
```
yulia-bel, line 429: const isProjected: boolean = year > lastAvailableDataYear

PR #745, opened 2022-11-03
Fixed an issue where scenarioId param was missing in table analysis

davidsingal:
Fixed an issue where the `scenarioId` param was missing when the user selects a scenario in the analysis table view.

1. Visit analysis, and go to table mode.
2. Choice an scenario from the sidebar
3. Ensure the `scenarioId` param is in the request of `impact/table` endpoint


elpamart:
LGTM!

PR #748, opened 2022-11-03
Children selection on intervention creation

JoaquinTrinanes:
### General description

This PR selects the leaf nodes of the business unit selector in the intervention creation form. It only applies to creation, not edition.

Creation example:
image

Edition example:
image

### Designs

### Testing instructions

## Checklist before merging

### DIFF
--- a/client/src/pages/data/scenarios/[scenarioId]/interventions/new.tsx
+++ b/client/src/pages/data/scenarios/[scenarioId]/interventions/new.tsx
@@ -54,7 +54,11 @@ const CreateInterventionPage: React.FC = () => {
       <div className="grid grid-cols-12 gap-6">
         <div className="col-span-10 col-start-2 xl:col-start-3 xl:col-span-8">
           <h1 data-testid="page-title">New intervention</h1>
-          <InterventionForm onSubmit={handleSubmit} isSubmitting={createIntervention.isLoading} />
+          <InterventionForm
+            onSubmit={handleSubmit}
+            isCreation
+            isSubmitting={createIntervention.isLoading}
+          />
         </div>
       </div>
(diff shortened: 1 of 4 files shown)


client/src/pages/data/scenarios/[scenarioId]/interventions/new.tsx
```diff
@@ -54,7 +54,11 @@ const CreateInterventionPage: React.FC = () => {
       <div className="grid grid-cols-12 gap-6">
         <div className="col-span-10 col-start-2 xl:col-start-3 xl:col-span-8">
           <h1 data-testid="page-title">New intervention</h1>
-          <InterventionForm onSubmit={handleSubmit} isSubmitting={createIntervention.isLoading} />
+          <InterventionForm
+            onSubmit={handleSubmit}
+            isCreation
```
Comment (davidsingal) [inline client/src/pages/data/scenarios/[scenarioId]/interventions/new.tsx:59]: It works, and It's ok. But did you consider instead of having `isCreation` something more explicit?? I'd like something like `selectCheckedStrategy="CHILD"`.

davidsingal (changes requested):
In addition to my comment, I miss the Changelog file update.

client/src/pages/data/scenarios/[scenarioId]/interventions/new.tsx
Comment (JoaquinTrinanes) [reply client/src/pages/data/scenarios/[scenarioId]/interventions/new.tsx:59]: I'd love to have that as a prop (something like `selectProps` or similar as you said). However, `checkedStrategy="CHILD"` only applies to the business unit selector, so a single prop like that wouldn't suffice.

If the select strategy has to be applied to ALL the selects, then that's another thing and I would gladly change it.

davidsingal (approved):

PR #751, opened 2022-11-04
Fix filters

JoaquinTrinanes:
### General description

This PR fixes the analysis filters sometimes clearing themselves.

When a scenario changed, we checked those filter to remove those that don't belong in the scenario, but didn't filter them recursively. Hence, selecting a child would invalidate the selection.

While another state solution isn't used, the `useQueryParam` hook was refactored to allow multiple observers in the same page. Otherwise, observing the current scenario without entering a infinite loop was next to impossible.

https://vizzuality.atlassian.net/browse/LANDGRIF-1021

## Checklist before merging


elpamart:
LGTM!

PR #765, opened 2022-11-14
Empty expanded tree parent

JoaquinTrinanes:
### General description

Refactor or tree selectors to solve some UI issues and improve usability.

https://vizzuality.atlassian.net/browse/LANDGRIF-1068
https://vizzuality.atlassian.net/browse/LANDGRIF-1030
https://vizzuality.atlassian.net/browse/LANDGRIF-1026

### DIFF
--- a/client/CHANGELOG.md
+++ b/client/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Update rows expanded state on analysis table [LANDGRIF-967](https://vizzuality.atlassian.net/browse/LANDGRIF-967)
 - Horizontal scroll on the legend [LANDGRIF-1082](https://vizzuality.atlassian.net/browse/LANDGRIF-1082)
 - Format of numbers of the contextual layers in the legend [LANDGRIF-1090](https://vizzuality.atlassian.net/browse/LANDGRIF-1090)
+- Tree Select search inconsistencies and items of depth 3 or deeper disappearing. [LANDGRIF-1026](https://vizzuality.atlassian.net/browse/LANDGRIF-1026) and [LANDGRIF-1030](https://vizzuality.atlassian.net/browse/LANDGRIF-1030)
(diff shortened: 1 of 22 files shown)


andresgnlez:
image

For some reason, when the user clicks on the filters and the dropdown displays on the analysis page, the **suppliers** selector has the focus by default. Not a big deal, but it's weird. This issue also happens in staging, which is unrelated to the latest code changes.

image


Last thing; when a user clicks on the right zone of the selector (around the chevron), the selector opens up but the focus is not set in the selector (despite having the focus-based styles applied) preventing the user to type in the box.

andresgnlez (changes requested):
image

For some reason, when the user clicks on the filters and the dropdown displays on the analysis page, the **suppliers** selector has the focus by default. Not a big deal, but it's weird. This issue also happens in staging, which is unrelated to the latest code changes.

image


Last thing; when a user clicks on the right zone of the selector (around the chevron), the selector opens up but the focus is not set in the selector (despite having the focus-based styles applied) preventing the user to type in the box.

davidsingal (changes requested):
Changelog is not updated

alexeh (commented):
I've seen some stuff that I am not sure it is right to left as is:

First of all, we should make sure that all the team is OK with the design of the 'search dropdown', since the font, its size.. changes:
image

Second, I realise that if you select a 'only son' through the dropdown, the parent gets shown in the field, however if you select it via search, the son gets shown:
image

image

Lastly, I know that the search match is flexible, but I am not sure if some results have any sense at all, check below:
image

image

I checked other behaviours and seems to work fine, but please @davidsingal @andresgnlez  
have a look at this.

I haven't checked for regressions, I will do a bit later

alexeh:
Also, @JoaquinTrinanes I realised this branch was originally taken from dev, but now points to staging. This can be potentially evil, there are some commits that are not part of this hotfix, so please be aware of that. Also, if its possible to squash them that would be great, so we can have a cleaner rebasing for everyone

PR #767, opened 2022-11-14
Replace lodash with lodash-es to decrease bundle size

JoaquinTrinanes:
### General description

This PR replaces `lodash` with `lodash-es`. The main difference is that the latter is written in ES Modules, allowing treeshaking.

The only way to use treeshaking with the normal library is to do the imports like `import x from lodash/x` instead of `import { x } from `lodash`. That's easy to get wrong and autocomplete is much worse in those cases.

A rule to forbid importing lodash (as it's a dependency of other packages) was added too.

## Checklist before merging


andresgnlez (approved):
LGTM 👍🏻

PR #794, opened 2022-11-25
fixed location type input for intervention form

davidsingal:
https://vizzuality.atlassian.net/browse/LANDGRIF-1089


alexeh (changes requested):
However this might not be within this PR´s scope, there is no way to test if it's working OK in other enviroments:

When editing you can't see which address (and I guess coords.) were introduced to generate this intervention. However the API provides that information:

image
image

You have to re-select the location type (which is shown first in lowe case, and after in capital case) in order the address/coords to be shown.
image

alexeh:
I also realise there are more cases not covered, I consider one of them, the first one, a bit critical:

1. When EDITING an existing scenario, if an Address was used, when a user wants to change this location using coordinates, those are not sent, it uses the previous address. In the example, I created an intervention using Barcelona as Address. Then, I want to edit the intervention using coordinates, but the address Barcelona is being sent:
image
Once the intervention is edited, you can check that Barcelona is still there.

I would suggest to mark this a a new bug, as probably we have no enough time to cover this with it's many use cases, we can discuss it later.

2. Validations:

Special characters are being validated as address or coordinates. I believe this should be validated on the FE side:
image


3. No feedback on incorrect input:
Even tho there is a message letting the user know: 
`Add lat and long coordinates separated by comma, e.g. 40, -3`

There are some uncovered use cases:
More than expected comma separated values are being validated as address
image

When CREATING a new intervention, if a comma is used for decimal separator, nothing happens when clicking the save button (not even the coordinates must be 90/180 validation), there is no feedback for the user, it can be confusing

image

alexeh (changes requested):
I've seen some bugs here:

1. An address won't get validated in the form for a Production aggreagation point location type
image

2. Even changing the location type, deleting the field and adding a new input, it won't get validated:
image

3. Changing to a location type that does not need an address/coordinates and pressing Save, will not send a request and will not show any feedback:
image


There are more bugs in EDITION of the intervention, but I believe we should focus our efforts on the CREATION part within this PR's scope

alexeh:
Issues found in EDITION:
1. Saving a intervention of location type Point of production sending same coordinates used in the CREATION, won't send any coordinates:
image
**NOTE: Adding a space between coords solves the issue**
**NOTE: this triggers an error in the API that we should cover as well, but the FE should send the full dto**

2. Having created a intervetion of type production aggregation point using coordinates, when changing JUST the location type to Point of production won't validate the coordinates:
image


3. Same happens if we introduce an Address:
image

4. Changing to any location type that does not need no address or coords, won't work and will give no feedback (related to first comment)
image

PR #834, opened 2023-01-19
Scenario action buttons should check roles and permissions of the user

mariasola:
- [x] Hook and example to check user role and permissions
- [x] Check permissions of the Scenario creation button in Analysis
- [x] Check permissions of the Scenario creation, edition, and deletion in the Admin > Scenarios page
- [x] Ensure an e2e test has been created
- [x] Update CHANGELOG


client/src/hooks/permissions/index.ts
```diff
@@ -0,0 +1,16 @@
+import { useProfile } from 'hooks/profile';
+
+import type { Permission } from './enums';
+
+export function usePermissions() {
+  const { data, isLoading } = useProfile();
+  const roles = data?.roles[0];
```
alexeh, line 7: Take in consideration that a user might have multiple roles with multiple permissions attached to each role

client/src/hooks/permissions/enums/index.ts
```diff
@@ -0,0 +1,12 @@
+/** User roles */
+export enum RoleName {
```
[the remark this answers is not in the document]
davidsingal: as we are not using a proper monorepo it's not so easy, as typescript doesn't read anything outside the __dirname we could do it throw an internal package or something like that. We'll check

client/src/containers/analysis-sidebar/component.tsx
```diff
@@ -14,10 +14,12 @@ import {
 } from 'store/features/analysis/scenarios';
 import { useInfiniteScenarios } from 'hooks/scenarios';
 import useBottomScrollListener from 'hooks/scroll';
+import { usePermissions } from 'hooks/permissions';
 import ScenariosFilters from 'containers/scenarios/filters';
 import { Anchor } from 'components/button';
 import Loading from 'components/loading';
 import ScenarioItem from 'containers/scenarios/item';
+import { Permission } from 'hooks/permissions/enums';
```
davidsingal, line 22:
```suggestion
import type { Permission } from 'hooks/permissions/enums';
```

client/src/hooks/permissions/index.ts
davidsingal: yes, roles should be an array

davidsingal (changes requested):

client/src/containers/analysis-sidebar/component.tsx
mariasola: @davidsingal we did not put a type in here because it is an `enum`, not a type. And the srict mode was telling us that setting a type there is an error.

client/src/hooks/permissions/enums/index.ts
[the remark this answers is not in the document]
alexeh: Mmmm right I see what you mean. I'm thinking we might add classes/enums/types as a node package in the root, or maybe tryout a oldschool rsync... I'll have a look as well!

client/src/hooks/permissions/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    role.permissions?.forEach((permission) => permissions.push(permission));
+  });
+
+  const hasRole = (role: RoleName) => {
+    return !isLoading && !!roles?.includes(role);
+  };
+
+  const hasPermission = (permission: Permission) => {
+    return !isLoading && !!permissions?.includes(permission);
+  };
```
alexeh, line 22: I think this wouldn't work, as you would be passing a enum value which would be transpiled to a string, and this is how [the permission entity looks](https://api.dev.landgriffon.com/swagger/#/User/UsersController_userMetadata) like.

Unless you specifically want the backend to retrieve permissions as an array of strings, happy to discuss that. But by default we will nest entities by its original shape.

Also maybe the naming on our side it's not the best? happy to discuss that as well

client/src/hooks/permissions/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  const permissions: Permission[] = [];
+
+  data?.roles?.forEach((role) => {
+    roles.push(role.name);
+    role.permissions?.forEach((permission) => permissions.push(permission));
+  });
+
+  const hasRole = (role: RoleName) => {
+    return !isLoading && !!roles?.includes(role);
+  };
```
alexeh, line 18: Take in account that unless you want this specifically to check for a admin role, this wouldn't have much sense, as we could not know what a role 'can do'. Unless for the time being you just want to discriminate between admin and user

client/src/containers/analysis-sidebar/component.tsx
```diff
@@ -141,6 +146,7 @@ const ScenariosComponent: React.FC<{ scrollref?: MutableRefObject<HTMLDivElement
                   className="block w-full"
                   variant="primary"
                   size="xl"
+                  disabled={!canCreateScenario}
```
alexeh, line 149: Right now we are not assigning any permission by default, should I quickly implement this on our side? It should be fairly quick happy to do so, let me know!

davidsingal: Yes, I think so. The roles have already permissions linked. So, they should be there. As we agreed, we will base the front end on the permissions we receive instead of asume the permissions by the role.

client/src/hooks/permissions/index.ts
davidsingal: Yes, we know. It will be there just in case we need for any reason. We will base the front end on the permissions instead.

client/src/hooks/permissions/index.ts
davidsingal: We know it couldn't work, it's like that because it's easier to work with a array of string and we can parse what we get from API. And last reason, because it's not documented the format of the permissions in Swagger and we thought it was still missing.

client/src/containers/analysis-sidebar/component.tsx
alexeh: Fair enough. On it!
Are we still aligned on our previous agreement that a admin role bypasses all permission checks?

client/src/hooks/permissions/index.ts
alexeh: All clear sir

client/src/containers/analysis-sidebar/component.tsx
davidsingal: Yes, but in other way. We'll allow unless permissions say no.

client/CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Styles of the legend were not applied correctly
 - Styles of the opacity control were not applied correctly
 
+### Added
+
+- New hook to check user role and permissions [LANDGRIF-1123](https://vizzuality.atlassian.net/browse/LANDGRIF-1123)
+
```
davidsingal, line 21: This part should be inside the #Unreleased section

client/CHANGELOG.md
```diff
@@ -20,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Styles of the legend were not applied correctly
 - Styles of the opacity control were not applied correctly
 
+### Added
+
```
davidsingal, line 28: This part is missing above

davidsingal (changes requested):

alexeh (changes requested):
@barbara-chaves hey thanks! Since you've added me as a reviewer, I'd like to point out some minor details. **Nothing related to FE code** as I am scared by CSS, DOM and most of the client-side technologies, but I see some weird stuff happening.

Please be aware that you've got [merge commits](6072a18733a56f09b4c2f53a5d1de407607f484f) that would be merged into the development branch, we should try to avoid that. 

There is another ['Rebase dev' commit](79f52dc83103c4fb8d1d3da484f36658dc8bf009) which we would also try to avoid as it is one of the points of rebasing, not having commits beyond our own in the history tree (among other things)

You can also see that there are BE commits (and changes) that already exists in the main dev branch meant to be "re-merged" as part of this PR, this can be potentially evil because of the different way GH treats rebase operation VS how git treats rebase operation. I want to highlight this last point because it has happened to us. more than once xD and ended up being a huge mess!!

If you are not sure how to proceed given our development workflow, or any concept in my feedback is new to you, please feel free to hit me up and I'll happily help you. I am not best example but I'll try me best to help :)

davidsingal (approved):

davidsingal (changes requested):
Screenshot 2023-02-08 at 11 39 41
I've created this scenario in the dev environment, the user should be able to edit it.

PR #842, opened 2023-01-30
Fixes in contextual layers

davidsingal:
* Avoid including no needed params to the contextual-layers endpoint, for example: `indicatorId`  https://vizzuality.atlassian.net/browse/LANDGRIF-1134
* Fix resolution to 4 https://vizzuality.atlassian.net/browse/LANDGRIF-1130


alexeh (changes requested):
Hey bro, there is something wrong here. Actually I cannot tell if this bug is introduced in this PR, or its actually in dev (can't reproduce it because of other bugs xD)

The "preview-contextual-layer" does not get applied to the main map, after you visualize it in the preview, and clicking apply.

Steps:

1. Open contextual layer model. Select any working map, Barley i.e. (we have some data issues)
2. Visualize (the map is requested at res 4 correctly
3. Apply. At this point you should see the contextual layer in the main map, but nothing happens

It's working OK in staging. **Again, since I can't reproduce the bug on dev, I'm adding this bug on this PR's scope**


image

alexeh:
Alright, in that case I would consider this PR as approved, feel free to merge it. Thanks!

alexeh (approved):

PR #864, opened 2023-02-13
[LANDGRIF-1173]: filtering indicators to show in chart and table

davidsingal:
Navigate to the chart and table, it should show only the indicators with status active

Task: https://vizzuality.atlassian.net/browse/LANDGRIF-1173


andresgnlez:
@davidsingal the UI has to display ALL indicators available and grey out those ones with `status === "inactive"`, so filtering them is not an option.

andresgnlez (approved):

alexeh (approved):

PR #873, opened 2023-02-16
Improve chart visualization

yulia-bel:
Update. Chart visualization implemented in 2 ways:

1. The default one - no parameter sent - for case of 'group by region':

- if no origin filters received, level 0 regions will be shown on chart (countries)
- if origin filters received and they are of level 0, their descendants of level 1 will be displayed
-  if origin filters received and they are of level 1, their descendants of level 2  will be displayed
-  if origin filters received and they are of level 2, the received regions will be displayed
-  if origin filters received and they are of different level, the chart will display regions and descendants of the lowest level received

2. Optional parameter 'depth' - can be applied for any group by entity. No matter the parent levels of received filters, user can choose the depth / level to be displayed on the chart

In order to use optional parameter FE changes are needed (add parameter to request)

Tests added.

### General description

### Designs

### Testing instructions

## Checklist before merging


api/src/modules/impact/dto/impact-table.dto.ts
```diff
@@ -161,6 +163,17 @@ export class GetRankedImpactTableDto extends BaseImpactTableDto {
   @IsUUID(4)
   scenarioId?: string;
 
+  @ApiPropertyOptional({
+    description:
+      'Optional property for groupBy region and chart view to choose level of admin regions to show',
```
alexeh, line 168: I think the msg is not correct as the grouping depends on the selected entity, so its not tied to 'admin-regions'

I would say something like:
"Select the depth of the tree for grouping elements, defaults to...."

api/src/modules/impact/dto/impact-table.dto.ts
```diff
@@ -161,6 +163,17 @@ export class GetRankedImpactTableDto extends BaseImpactTableDto {
...
+  })
+  @IsOptional()
+  @Type(() => Number)
+  @IsNumber()
+  @Min(1)
+  @Max(3)
```
alexeh, line 174: I think the minimum should be 0, as discussed. If a user wants to select 6 countries and see a 4 ranked still at country level, that depth should equal 0

Also, I wouldn't set a fixed max limit. Even tho the current limit makes sense, I think its a better idea to not set a max limit and just uncollapse the tree as much as it can be uncollapsed.

i.e, users sends a 5 value but we have a maximum depth of 3, just send back a 3 level deep tree

api/src/modules/impact/dto/impact-table.dto.ts
```diff
@@ -161,6 +163,17 @@ export class GetRankedImpactTableDto extends BaseImpactTableDto {
...
+  level?: number;
```
alexeh, line 175: to be more consistent with other endpoints, I would name this parameter as 

```depth```

File: api/src/modules/impact/impact.service.ts [hunk 4]
```diff
@@ -144,6 +144,21 @@ export class ImpactService extends BaseImpactService {
       entities,
     );
 
+    // If chart is requested for regions and specific level is received, we get only the requested level from admin regions trees of the response
+
+    if (
+      rankedImpactTableDto.groupBy === 'region' &&
```
alexeh, line 150: Following https://github.com/Vizzuality/landgriffon/pull/873/files#r1109319576, I still dont get why this is somehow tied to regions.

Does not make sense to apply this behaviour to any other entity that is a tree structure?

api/src/modules/impact/impact.service.ts
```diff
@@ -465,4 +480,27 @@ export class ImpactService extends BaseImpactService {
       isProjected: false,
     };
   }
+
+  private getLevelOfImpactTableRows(
+    impactTableRows: ImpactTableRows[],
+    level: number,
+  ): ImpactTableRows[] {
+    if (level === 1) {
```
alexeh, line 488: This should be 0, in order to uncollapse the whole thing

api/src/modules/impact/impact.service.ts
```diff
@@ -465,4 +480,27 @@ export class ImpactService extends BaseImpactService {
...
+      return impactTableRows;
+    }
+    if (level < 1) {
```
alexeh, line 491: I think this is not necessary, because:
1. first level should be 0, which is a valid value
2. there should be a default behaviour to default to

api/src/modules/impact/impact.service.ts
```diff
@@ -465,4 +480,27 @@ export class ImpactService extends BaseImpactService {
...
+      (impactTableRows: ImpactTableRows[], row: ImpactTableRows) => {
+        if (row.children) {
+          impactTableRows = impactTableRows.concat(
+            this.getLevelOfImpactTableRows(row.children, level - 1),
+          );
+        }
+        return impactTableRows;
+      },
+      [],
+    );
```
alexeh, line 504: This is nice :)

However, maybe we should try to use immutability, thoughts?

api/test/e2e/impact/mocks/chart-levels-preconditions/chart-levels.preconditions.ts
```diff
@@ -0,0 +1,132 @@
...
+    municipalityThree,
+    municipalityFour,
+    municipalityFive,
+    municipalitySix,
+    municipalitySeven,
+  ];
+  const sourcingLocations: SourcingLocation[] = [];
+
+  for (const municipality of municipalities) {
+    const i: number = municipalities.indexOf(municipality);
```
alexeh, line 110: This is a super lightweight improvement, but If you realise, for each element on municipalities, you are running through municipalities again to find the index.

You could either get the index in the for of loop, or use a native array method that can provide you the index as one of the parameters in its callback

api/test/e2e/impact/impact-chart/chart-levels.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+        .expect(HttpStatus.OK);
+
+      expect(response.body.impactTable[0].rows.length).toBe(5);
+      expect(response.body.impactTable[0].rows[0].name).toBe('Municipality 1');
+      expect(response.body.impactTable[0].rows[0].values[0].value).toBe(1000);
+      expect(
+        response.body.impactTable[0].others.aggregatedValues[0].value,
+      ).toBe(2000);
+    },
+  );
```
alexeh, line 108: Nice, very clean, creating preconditions just once etc...

api/test/e2e/impact/impact-chart/chart-levels.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+  let jwtToken: string;
+  let dataSource: DataSource;
+  let indicator: Indicator;
+
+  beforeAll(async () => {
+    testApplication = await ApplicationManager.init();
+
+    dataSource = testApplication.get<DataSource>(DataSource);
+
+    ({ jwtToken } = await setupTestUser(testApplication));
```
alexeh, line 23: This is cool, I had no idea you could assign a value like this

PR #879, opened 2023-02-22
Data/feature/update deforestation pipe mask leviv

aagm:
### General description

This PR is to provide the masking capabilities on managed forests to the data pipes for deforestation and forestGHG.

## Checklist before merging


.gitignore
```diff
@@ -4,6 +4,7 @@ node_modules
 /api/dist
 /api/coverage
 .env
+.vscode
```
alexeh, line 7: we better not merge this or someone will hunt us with a stick

data/h3_data_importer/Makefile
```diff
@@ -332,23 +332,47 @@ download-deforestation-list:
...
 	table_name=`basename -s .tif $(FILE) | tr -d ' \t\n\r' | grep -oP "\d{2}[A-Z]_\d{3}[A-Z]"`; \
+	mask_path=`$${table_name}_mask`; \
+	mask_table_name=`basename -s .tif $${mask_path}; \
  	wget -q -O "$(WORKDIR_DEFORESTATION)/$${table_name}.tif" $(FILE); \
 	rm -f "$(WORKDIR_DEFORESTATION)/Hansen_GFC_2020_lossyear/Loss_$${table_name}_10km.tif"; \
-	gdal_calc.py --quiet --calc "(A>19)" --format GTiff --co NBITS=1 --co compress=PACKBITS --co tiled=YES --co BLOCKXSIZE=400 --co BLOCKYSIZE=400  --NoDataValue 0 --overwrite \
+	rm -f "$(WORKDIR_DEFORESTATION)/$${mask_table_name}_clip.tif"; \
+	rm -f "$(WORKDIR_DEFORESTATION)/$${mask_table_name}.tif"; \
+	rio clip "$${mask_path}" "$(WORKDIR_DEFORESTATION)/$${mask_table_name}_clip.tif" \
+		--like "$(WORKDIR_DEFORESTATION)/$${table_name}.tif"; \
```
BielStela, line 368: This clips the whole raster of mask for managed forests to the extend of the Hansen deforestation tile. I'm getting this right @aagm ?

data/h3_data_importer/Makefile
```diff
@@ -332,23 +332,47 @@ download-deforestation-list:
...
+	gdalwarp -q -s_srs EPSG:4326 -t_srs EPSG:4326 -r nearest -tr 0.00025 0.00025 -multi -of GTiff  -ot UInt32 -wo NUM_THREADS=ALL_CPUS \
+		'$(WORKDIR_DEFORESTATION)/$${mask_table_name}_clip.tif' \
+		'$(WORKDIR_DEFORESTATION)/$${mask_table_name}.tif'; \
+	gdal_calc.py --quiet --calc "(A>16)*(B<12)" \
+		--format GTiff --co NBITS=1 --co compress=PACKBITS --co tiled=YES --co BLOCKXSIZE=400 --co BLOCKYSIZE=400  --NoDataValue 0 --overwrite \
 		-A "$(WORKDIR_DEFORESTATION)/$${table_name}.tif" \
+		-B "$(WORKDIR_DEFORESTATION)/$${mask_table_name}.tif" \
```
BielStela, line 375: Pixel values less than 12 in the Lesiv dataset mean only category 11 - Naturally regenerating forest without any signs of human activities, e.g., primary forests.
Wouldn't it make sense to also include "20 - Naturally regenerating forest with signs of human activities, e.g., logging, clear cuts etc." since are natural too? (just thinking out loud)
For example, would this rule out all the deforestation events in the tipical spanish "monte" where the forest is kinda natural ( in the sense that is not a plantation, and that has a great ecosystemic value) but has some human intervention? (maybe I should just read the Lesiv paper and review the methodology )

data/h3_data_importer/Makefile
```diff
@@ -337,6 +337,14 @@ download-deforestation: download-deforestation-list
...
 	cat $(WORKDIR_DEFORESTATION)/Hansen_GFC_2020_lossyear.txt | xargs -P $(PARALLELIZATION_FACTOR) -I {} /bin/bash -c 'make transform-deforestation-file FILE={}'
 
+# Managed forest Mask from Lesiv(2022) - to be used with deforestation and forest ghg.
+# data download from https://zenodo.org/record/5879022
+#After downloading the mask layer,
+download-managed-forests:
+	@echo "Downloading managed forests file... "
+	mkdir -p $(WORKDIR_DEFORESTATION)/Lesiv_2015_managed_forests
+	wget -q -O $(WORKDIR_DEFORESTATION)/Lesiv_2015_managed_forests/managed_forests.tif "https://zenodo.org/record/5879022/files/FML_v3-2_with-colorbar.tif?download=1"
+
```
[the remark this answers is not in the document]
aagm: yes, this is something we discuss to start doing at some point but i think it should be a call for @elpamart  and not the aim of this pr

data/h3_data_importer/Makefile
aagm: yes, you are getting it right, so we avoid generating the warped of the full raster, that could be, putting it into soft words, problematic

data/h3_data_importer/Makefile
aagm: Well the authors follow the [FAO 2015 FRA definitions](https://www.fao.org/3/ap862e/ap862e00.pdf), (the latest one from 2020 can be found [here](https://www.fao.org/3/I8661EN/i8661en.pdf))
The 2015 definition can be read as:
> OTHER NATURALLY REGENERATED FOREST
> Naturally regenerated forests where there are clearly visible indications of human activities.
> Explanatory notes
> 1. Includes selectively logged-over areas, areas regenerating following agricultural land use, areas recovering from human-induced fires, etc.
> 2. Includes forests where it is not possible to distinguish whether planted or naturally regenerated.
> 3. Includes forests with a mix of naturally regenerated trees and planted/seeded trees, and where the naturally regenerated trees are expected to constitute more than 50 percent of the growing stock at stand maturity

Note that this definition has been erased from FAO 2020 FRA.

Then from the authors,  

> We have calculated and compared the areas of mapped forest management classes with official FAO FRA statistics for the year 2015[37](https://www.nature.com/articles/s41597-022-01332-3#ref-CR37).... Specifically, naturally regenerating forests (as referred to in the FAO FRA), correspond to our classes 11 and 20, and planted forests (as called in the FAO FRA) correspond to the mapped classes 31 and 32. The only exception is semi-natural forests, which belong to planted forests in the FAO FRA.

So a priori, you are right, but then, from the authors, we also have this:

> From the confusion matrix, it follows that planted forests (class 31) are, unsurprisingly, underestimated. This class is confused with naturally regenerated managed forests (20). The confusion matrix also indicates that there should be double the amount of planted forests that are currently shown on the map.

Also taking the definition from FAO 2020 FRA of what deforestation means:

> DEFORESTATION
> The conversion of forest to other land use independently whether human-induced or not.
> Explanatory notes
> 1. Includes permanent reduction of the tree canopy cover below the minimum 10 percent threshold.
> 2. It includes areas of forest converted to agriculture, pasture, water reservoirs, mining and urban areas.
> 3. The term specifically excludes areas where the trees have been removed as a result of harvesting or logging, **and where the forest is expected to regenerate naturally or with the aid of silvicultural measures.**
> 4. The term also includes areas where, for example, the impact of disturbance, over-utilization, or changing environmental conditions affects the forest to an extent that it cannot sustain a canopy cover above the 10 percent threshold.

My rationale is to be strict and consider class 11 only, this is because it represents primary forest, and we will be representing deforestation risk over not significant disturbed areas. 
We can discuss taking class 20 into consideration, but in terms of what it means, we will be overestimating deforestation risk, and the definition of deforestation, might not be accurately implemented. 

In terms of numbers, this will have a deep impact. Globally,  **class 11** represents approx 40% of all forested areas while **class 20** is approx 34%, and the **plantation classes (31, 32, 40, 53)** are 25%. (note as well that these numbers differ from FAO FRA ones, and this is also warn by the authors)

BielStela: Hmmm  quite a soup of definitions of what is and what is not deforestation.., so wrapping things up, if we use class 20 we will be overestimating deforestation, and underestimating if we don't use it. I'm allright?

On the other hand, I don't understand 
> _and where the forest is expected to regenerate naturally or with the aid of silvicultural measures_

Does it mean that if i.e.  I chop of 1 square km of rain forest and it restores itself naturally in 100 years it is not considered deforestation?

BielStela: Maybe @mikeharfoot  has a word or two about this :smiley: ?

aagm: What FRA's definition means, is that if there is a sign of management in that patch, we cannot call it deforestation (or I would add **illegal** there), as any disturbance observed there, will be because of that management. Take into account that deforestation is a change in land use, so if it was forest, it was lost but is later regenerated and there is not a change in its land use categorization we cannot define it as deforestation

.gitignore
alexeh, line 7: Dunno if this never left, (because the thread is marked as resolved) but it's still here

BielStela: @aagm can you remove this before we merge?

BielStela (approved):

PR #880, opened 2023-02-22
Test specs re-organization

davidsingal:
* Merging all analysis specs in one file
* Centralizing all the intercepts in a command called `interceptAllRequests`


andresgnlez:
On the first run, I run into some failed tests... This might be because I removed the `.next` folder and compilation times might take longer than usual. So I run tests a couple of times more: the second run failed as well but the third run and so on went well and all specs passed.

With this... Not sure what I want to mean, but that was my experience hahaha. I want to believe it was "bad luck", anyway, I will run tests again here in the PR to be sure :)

andresgnlez (approved):

barbara-chaves (approved):

PR #894, opened 2023-02-27
[LANDGRIF-1249]: Wrong behavior in year selector

barbara-chaves:
### General description

This PR fixes the behaviour of the year selector filter on analysis/table

### Testing instructions

- The default last year selected should be the last year with data + 5 years;
- The selectors should allow search values;
- The 'end' year should accept any valid year value bigger then the 'start' year;

## Checklist before merging

### DIFF
--- a/client/src/components/forms/select/autocomplete/component.tsx
+++ b/client/src/components/forms/select/autocomplete/component.tsx
@@ -113,9 +113,6 @@ const AutoCompleteSelect = <T,>({
               })}
             >
               {filteredOptions[index].label}
-              {filteredOptions[index].extraInfo && (
-                <span className="text-sm">{` - ${filteredOptions[index].extraInfo}`}</span>
-              )}
             </div>
           )}
         </Combobox.Option>
@@ -162,7 +159,7 @@ const AutoCompleteSelect = <T,>({
               </Combobox.Label>
             )}
             <div className="relative w-full overflow-hidden">
-              <Combobox.Input<'input', Option>
+              <Combobox.Input<'input', Option<T>>
                 className={classnames(
                   'w-full inline-flex items-center py-2.5 pl-3 pr-10 text-left leading-5',
                   'bg-white border rounded-md shadow-sm cursor-default hover:cursor-pointer focus:border-navy-400 focus:outline-none focus:ring-0',
--- a/client/src/components/forms/select/autocomplete/component.tsx
+++ b/client/src/components/forms/select/autocomplete/component.tsx
@@ -29,6 +29,7 @@ const AutoCompleteSelect = <T,>({
   clearable = false,
   onChange,
   onClearSelection,
+  onSearch,
   ...props
 }: AutoCompleteSelectProps<T>) => {
   const [virtualizedListWidth, setVirtualizedListWidth] = useState(0);
@@ -69,8 +70,11 @@ const AutoCompleteSelect = <T,>({
   );
 
   const handleInputChange = useCallback(
-    (event: ChangeEvent<HTMLInputElement>) => setQuery(event.target.value),
-    [],
+    (event: ChangeEvent<HTMLInputElement>) => {
+      setQuery(event.target.value);
+      onSearch?.(event.target.value);
+    },
+    [onSearch],
   );
 
   const handleAfterTransitionLeave = useCallback(() => setQuery(''), []);
@@ -109,6 +113,9 @@ const AutoCompleteSelect = <T,>({
               })}
             >
               {filteredOptions[index].label}
+              {filteredOptions[index].extraInfo && (
+                <span className="text-sm">{` - ${filteredOptions[index].extraInfo}`}</span>
+              )}
             </div>
           )}
         </Combobox.Option>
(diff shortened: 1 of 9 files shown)


client/src/components/forms/select/autocomplete/types.d.ts
```diff
@@ -16,4 +16,5 @@ export type AutoCompleteSelectProps<T = string> = Omit<
   rowHeight?: number;
   clearable?: boolean;
   onClearSelection?: () => void;
+  onSearch?: (value: string) => void;
```
andresgnlez, line 19: `(value: T) => void;` better, so we can ensure the returned value is of the same type as we set in the options.

andresgnlez: Anyway, why do you need this `onSearch` callback for? 🤔

client/src/containers/analysis-visualization/analysis-filters/years-range/component.tsx
```diff
@@ -23,37 +27,54 @@ const YearsRange: React.FC = () => {
 
   const { startYear, endYear, yearsGap, setYearsRange } = useYearsRange({
     years,
+    yearsGap: 1,
     // Map mode only makes use of the endYear and will display the Select,
     // not the YearsRangeFilter.
-    validateRange: visualizationMode !== 'map',
+    validateRange: visualizationMode !== 'map' && visualizationMode !== 'table',
```
andresgnlez, line 33: Isn't it easier to just validate when `visualizationMode === 'chart'` ?

client/src/containers/filters/years-range/component.tsx
```diff
@@ -52,31 +52,29 @@ export const YearsRangeFilter: React.FC<YearsRangeFilterProps> = ({
...
+  //   onChange?.({ startYear: Number(startYearOption.value), endYear: Number(endYearOption.value) });
+  // }, [
+  //   startYear,
+  //   endYear,
+  //   startYearOptions,
+  //   endYearOption,
+  //   endYearOptions,
+  //   onChange,
+  //   startYearOption,
+  // ]);
```
andresgnlez, line 77: Do not leave uncommented code.

client/src/containers/analysis-visualization/analysis-filters/years-range/component.tsx
andresgnlez: Anyway, why are we excluding the `table` mode of the validation now? This was not the previous behaviour, wasn't it?

client/src/containers/filters/years-range/component.tsx
```diff
@@ -94,6 +92,10 @@ export const YearsRangeFilter: React.FC<YearsRangeFilterProps> = ({
   // Prevent display when not loaded
   if (!isLoaded) return null;
 
+  const handleChangeEndYear = ({ value }) => {
+    onChange?.({ startYear, endYear: Number(value) });
```
andresgnlez, line 96: Aren't we passing the options as `number` ? We shouldn't need to parse the value at this endpoint, it should be a `number` already.

client/src/components/forms/select/autocomplete/component.tsx
```diff
@@ -109,6 +113,9 @@ const AutoCompleteSelect = <T,>({
               })}
             >
               {filteredOptions[index].label}
+              {filteredOptions[index].extraInfo && (
+                <span className="text-sm">{` - ${filteredOptions[index].extraInfo}`}</span>
+              )}
```
andresgnlez, line 118: I don't think this solution scales properly, we are assuming here a way to pass extra info to the value but given a certain format (` - {whatever}`), but what if I want to add info with a different format? I just can't.

Anyway, why don't just modify the label of the options?

barbara-chaves: I just used the same property being used before (extraInfo), but I like your solution better. I will change it.

client/src/containers/analysis-visualization/analysis-filters/years-range/component.tsx
barbara-chaves: I was testing a solution and add it, then forgot to remove.

client/src/containers/filters/years-range/component.tsx
```diff
@@ -148,23 +130,21 @@ export const YearsRangeFilter: React.FC<YearsRangeFilterProps> = ({
...
-                  onChange={({ value }) => {
-                    onChange?.({ startYear: Number(value), endYear });
-                  }}
+                  onChange={({ value }) => onChange?.({ startYear: value, endYear })}
                   placeholder={placeholderFrom}
                 />
                 <div>To</div>
                 <Select<number>
                   loading={loading}
                   options={endYearOptions.map((option) => ({
```
andresgnlez, line 139: Can we move the options to a `useMemo` hook, please? So we avoid mapping them in every render.

andresgnlez:
@barbara-chaves to put the icing on the cake, let's add some tests testing the new behaviors so we ensure we have no regressions.

davidsingal:
This PR should point to staging, could you change it? @barbara-chaves

PR #898, opened 2023-02-28
(chore) tiler general improvements

alexeh:
### General description

This PR adds some little improvements and a first version docs to the LandGriffon Tiler.

I think I should document the workflow of the microservice, not sure where should I do so.

Happy to discuss everything


infrastructure/kubernetes/modules/env/main.tf
```diff
@@ -146,9 +146,13 @@ module "k8s_tiler" {
     {
       name = "DEFAULT_COG"
       value = "biomass.tif"
+    },
+    {
+      name = "REQUIRE_AUTH"
+      value = "false"
```
alexeh, line 152: I am setting this as false (for now) because:

1. Adding a auth method in FastAPI's swagger apparently is not as trivial as it is in other frameworks
2. Science team will likely be testing this through swagger until we get the right configuration
3. The cog served by the tiler is a generic one for testing purposes

tiler/pyproject.toml
```diff
@@ -0,0 +1,14 @@
...
+pythonpath = [
+  "."
+]
+env = ["REQUIRE_AUTH=true",
+    "API_HOST=",
+    "API_PORT=3000",
+    "S3_BUCKET_NAME=mybucket",
+    "ROOT_PATH=",
+    "TITILER_PREFIX=/cog",
+    "TITILER_ROUTER_PREFIX=/cog"]
```
alexeh, line 14: I explored different options for setting test env vars in python, I decided to add this, even tho we can inject them, I think we should provide a way to have some test env vars as we do in a node config package, instead of adding a env.test file

tiler/app/middlewares/url_injector.py
```diff
@@ -31,7 +31,7 @@ def s3_presigned_access(url: str | None = Query(default=None, description="Optio
     """
     if not url:
         if not default_cog:
-            raise Exception("DEFAULT_COG env var is not set. It is required if no URL is paased to the tiler")
+            raise Exception("DEFAULT_COG env var is not set. It is required if no URL is passed to the tiler")
```
BielStela, line 34: a little bit nit pick but this could be `ValueError` instead of a plain `Exception`

tiler/pyproject.toml
BielStela: hmm don't know, never seen anything like this before for testing :shrug:. Up to you!

tiler/README.md
```diff
@@ -0,0 +1,27 @@
...
+## API Documentation
+
+Once the Tiler service is running, the API documentation can be accessed at `http://localhost/tiler/docs`.
+
+
+## TODO
+
+- Add more env vars that are required, for deployment and fine-tuning
+- Add more tests (there is a basic pipeline set up with a couple of test)
+- Investigate caching options to improve performance
```
tiagojsag, line 27: every time you cache for performance, god kills a kitten

tiagojsag (approved):

PR #899, opened 2023-02-28
[LANDGRIF-1262] Refactor (removes) `tiff_to_h3_table.py` to something more modular

BielStela:
### General description

Rewrite of the raster to h3 ingestion. Highlights: 
- Updates to  use of `psycopg` 3 api
-  Removes some archeological remains (on the way to the British museum)
- overall rewrite the file with love, trying to make it easy to follow  and modular to separate DB things from raster->h3. 

### Testing instructions

:warning: **TODO** THIS MUST BE REVIEWED AND CHECKED CAREFULLY

- Make sure it can handle the bigger imports like Hansen :heavy_check_mark: 
- Is _FAST_ enought :yellow_circle: 
- Yields exacly (or maybe better) db state as the old script  
- Everyone likes it

### Known issues & a bit of Discussion
The process has 3 bottlenecks currently _[seconds running SPAM ingestion]_: 
1) Convert a list of raster files (as many as there are in the folder parameter) to h3 index -> value series. This has been solved by using multiprocess pool since is quite embarrassingly parallel. [24s]
2) Join the resulting Series/Dataframes into a big fat DataFrame to use as source for the DB. This is slow and I'm already (I think) using a fast way of joining dataframes using the h3 index as index and `DataFrame.join(list[DataFrame])`[57s]
3) The DB ingestion is also single threaded and writing the dataframe to the h3_grid_table takes a while for SPAM [34s]

So for the SPAM case, which has 51 rasters to ingest (the worst case), the old script takes ~90s and the new approach takes 116s. 

For datasets with a single raster it is faster. 

All this has to be put into perspective that the longest time in the ingestion is taken by the deforestation (hansen + ghg) part which is 99.9% downloading and preprocessing with gdal and the resulting raster ingested is as small as a single SPAM one. 
Thus, the next big step should be move the postprocessed results of deforestation to the cloud.

And why you may ask? I find it a little bit less cluttered and easy to follow to apply modifications safely.

 PD.: _I almost forget, this PR also removes unused files in the data folder which are quite annoying to have around_

## Checklist before merging


alexeh (approved):
Lemme know when we can fire this up to test :)

PR #900, opened 2023-02-28
[LANDGRIF-1169]: Change impact layer resolution according to zoom value

barbara-chaves:
### General description

This PR changes the resolution of the H3 impact layer to increase based on the map zoom value.

When there is a location filter applied:

- In zoom less than 5 it should always render resolution 4
- In zoom between 5 to 7 it should render resolution 5
- In zoom greater than 7 it should render resolution 6

### Testing instructions

Go to analysis/map
Click the zoom in button several times. The map should not be refetched.
Restore the zoom to the minimum value.

Select one or more 'location type' filters. Click the zoom in button 3 times. The map should refetch
(The initial map zoom is 2, so with 3 clicks it goest to zoom 5 and triggers the fetch of impact layer with resolution 5)
Click 3 more times and it should refetch again, now with resolution 6.

Do the inverse process to check if it fetches when the zoom is reduced. 
You can check the payload of the endpoint fetching the layer on network searching for the endpoint '/h3/map/impact'

## Checklist before merging


client/package.json
```diff
@@ -78,6 +78,7 @@
     "rooks": "5.10.0",
     "sharp": "^0.31.1",
     "tailwindcss": "3.2.4",
+    "use-debounce": "^9.0.3",
```
davidsingal, line 81: if I'm not wrong rooks package already includes useDebounce, could you check it?

davidsingal (changes requested):
It looks good to me, I've only left a comment about a package that could be duplicated

davidsingal (approved):

davidsingal:
could you check with Andrés if you will have some conflicts with the layer manager integration here? #903

davidsingal:
Please use the origin filter instead of location type @barbara-chaves

client/src/components/map/component.tsx
```diff
@@ -71,14 +72,29 @@ const Map = React.forwardRef<DeckGLRef, MapProps>(
...
+    );
+
+    const handleMapMove = useCallback(
+      (_viewState: Parameters<MapProps['onViewStateChange']>[0]) => {
+        setLocalViewState(_viewState.viewState);
+        debouncedViewStateChange(_viewState);
+      },
+      [debouncedViewStateChange],
+    );
+
```
davidsingal, line 90: I think this is the reason why moving the map with the mouse it very slow

davidsingal (changes requested):

andresgnlez (approved):

davidsingal (approved):

PR #903, opened 2023-03-02
[LANDGRIF-1248]: local layer manager integration

andresgnlez:
### General description

This PRs brings, mainly two changes:
- DeckGL's map is deprecated in favor of `react-map-gl` (this change also fixes issues related to the trackpad in the current map).
- Implementation of a local layer manager that handles, so far, DeckGL-kind layers but could handle, eventually more types of layers: https://deck.gl/docs/api-reference/layers

https://vizzuality.atlassian.net/browse/LANDGRIF-1248
https://vizzuality.atlassian.net/browse/LANDGRIF-1254

_My main concern is we need to disable React's strict mode to make this approach of local manager works. This is not desirable, and, although it might work now, I would recommend taking a look again in the future to try to enable it again._

Also, I tested this PR in **Chrome**, **Firefox** and **Safari** and works fine.

### Testing instructions

- `yarn` to install/update/remove dependencies.
- go to `/analysis/map` and play adding, sorting, and removing layers. Also check interactivity (tooltip) displays correct information even when layers are reordered.
- Also check the preview map (the one in the modal) works as expected.

## Checklist before merging


andresgnlez:
We _might_ not need the Layer Manager library after all, on hold until further notice.

davidsingal:
I've tried and it works perfectly. Once you cleaned up the PR let us know to update the Jira tasks.

davidsingal (approved):

PR #905, opened 2023-03-06
Fix update user roles

alexeh:
### General description

This PR fixes a bug where roles where not updating for a user.

The roles are sent as a array of strings which has to be converted to a proper entity before saving them to the database.

Using a basic update statement by typeorm repository, the entity will get updated by the values that matches the entity, and it will return the same DTO that has consumed, meaning that, during updating, the array of strings `roles` will be ignored, but it will be returned

### Designs

### Testing instructions

Given a admin role, update any users roles via PATCH /user.
It will return the roles as entities in the role property in the user object, plus some permissions associated. if any role has any

## Checklist before merging


yulia-bel (approved):

api/src/modules/users/users.service.ts
```diff
@@ -183,4 +183,17 @@ export class UsersService extends AppBaseService<
   async deleteUser(userId: string): Promise<DeleteResult> {
     return this.repository.delete(userId);
   }
+
+  async updateUser(userId: string, updateUser: any): Promise<User> {
```
alexeh, line 187: set correct typing

api/src/modules/authentication/authentication.service.ts
```diff
@@ -282,4 +284,8 @@ export class AuthenticationService {
       throw new ConflictException('Email already exists.');
     }
   }
+
+  updateRoles(rolesEnum: ROLES[]): Role[] {
```
alexeh, line 288: better naming:
get role entities
create roles from names

PR #951, opened 2023-05-16
change metadata rounding from 2 to 4

elpamart:
### General description

This PR explores the visualization that is reached when modifying the round of the quantiles obtained by the BE in the impact map.  The previous round applied (2 decimal places) did not provide the best visualization when filtering the data by a specific location as the difference end up being really small that is not appreciated with that level of detail (desegregation of the h3 from res4 to res6).

As has been commented on the jira task, there are different kinds of visualizations that can be applied on this case:

- continue with the quantiles but increase the granularity for obtaining a better visualization (from 2 to 4 decimal places)
- explore the application of jenks natural breaks. 

At this point we have done a quick fix that consist in continuing with the percentiles categorization but changing from 2 to 4 decimal places. The percentiles categorization provides a straightforward and intuitive way to understand the distribution of data, allowing for an easy identification of the relative position of values within the dataset. Even though the natural breaks could provide an improved visualization, its application is more complex and it can be more sensitive to outliers. Based on this, before starting with the application of this type of visualization (a task has been already created in jira), we need to minimize the number of outliers that our calculations may present. 

NOTE: We have not considered the equal intervals visualization as this type of categorization does not consider the distribution or patterns in the data, potentially resulting in inadequate representation of the underlying structure. 

### Testing instructions

Explore how the local visualization of the impact map has changed vs the data that is provided in the dev environment (especially when we filter by an specific location). 

## Checklist before merging

### DIFF
--- a/api/src/modules/h3-data/h3-data.repository.ts
+++ b/api/src/modules/h3-data/h3-data.repository.ts
@@ -785,7 +785,7 @@ export class H3DataRepository extends Repository<H3Data> {
     // If map type is Impact Map, sum up values from previous subquery
     if (mapType === IMPACT_MAP_TYPE.IMPACT_MAP) {
       impactViewAggregationQueryBuilder.addSelect(
-        'round(sum(impactview.value * reduced.scaled_value)::numeric, 2)',
+        'round(sum(impactview.value * reduced.scaled_value)::numeric, 4)',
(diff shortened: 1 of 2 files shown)


alexeh (approved):
congrats on understanding how this query is built!

PR #984, opened 2023-06-14
Add download of upload file errors

barbara-chaves:
### General description

This PR adds the download of the upload errors

### Testing instructions

1. Upload a file with errors
2. After the uploading process is finished, an error message will be displayed and a download button should be enabled.
3. Click on the button. A csv file with 2 columns (line, error) should be downloaded.   

### Related task
https://vizzuality.atlassian.net/browse/LANDGRIF-1342

## Checklist before merging

### DIFF
--- a/client/src/hooks/tasks/index.ts
+++ b/client/src/hooks/tasks/index.ts
@@ -1,6 +1,6 @@
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 
-import { apiService } from 'services/api';
+import { apiRawService, apiService } from 'services/api';
 
 import type { UseQueryResult, UseQueryOptions, UseMutationResult } from '@tanstack/react-query';
 import type { APIMetadataPagination, ErrorResponse, Task } from 'types';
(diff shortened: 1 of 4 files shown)


client/src/hooks/tasks/index.ts
```diff
@@ -100,3 +100,27 @@ export function useUpdateTask(): UseMutationResult<TaskAPIResponse> {
...
+          },
+        })
+        .then(({ data: responseData }) => responseData),
+    {
+      ...DEFAULT_QUERY_OPTIONS,
+      ...options,
+    },
+  );
+  return query;
+}
```
davidsingal, line 126:
```suggestion
export function useDownloadTaskErrors(
  id: Task['id'],
  options: UseQueryOptions<string> = {},
): UseQueryResult<string> {
  const query = useQuery<string>(
    ['task-error', id],
    () =>
      apiRawService
        .request({
          method: 'GET',
          url: `/tasks/report/errors/${id}`,
          params: {
            type: 'sourcing_data_import',
          },
        })
        .then(response => response.blob())
        .then(blob => {
            var url = window.URL.createObjectURL(blob);
            var a = document.createElement('a');
            a.href = url;
            a.download = "filename.xlsx";
            document.body.appendChild(a); // we need to append the element to the dom -> otherwise it will not work in firefox
            a.click();    
            a.remove();  //afterwards we remove the element again         
        }),
    {
      ...DEFAULT_QUERY_OPTIONS,
      ...options,
    },
  );
  return query;
}
```

This is another way to make it downloadable. I like this way because it doesn't parse the data, it just downloads from API. So, it has better performance.

davidsingal (changes requested):

client/src/hooks/tasks/index.ts
barbara-chaves: It didn’t work well for me, so I am using the  previous way

davidsingal (approved):

PR #1016, opened 2023-07-10
Add edit modal and delete user

barbara-chaves:
### General description

This adds the 'edit' button to the user list, with edit the form disabled and the 'delete user' functionality working
It changes the users list to don't display the current user name

It updates the users page style and table content 

### Designs

[users list](https://www.figma.com/file/7bmF29CIXDJwCpuloW5Nt6/%F0%9F%94%B5-Landgriffon---Redesign_BLUE?type=design&node-id=1870-167404&mode=dev)
[edit user modal](https://www.figma.com/file/7bmF29CIXDJwCpuloW5Nt6/%F0%9F%94%B5-Landgriffon---Redesign_BLUE?type=design&node-id=1870-169568&mode=dev)

### Testing instructions

1. With an admin account, go to '/profile/users'
2. the list should NOT display your current user name
3. click on some 'edit' button on the list. A modal should open with the edit user form and the 'save' button should be disabled
4. At the bottom of the list are some test users to be deleted. Try to delete them.

### Tasks
https://vizzuality.atlassian.net/browse/LANDGRIF-1346
https://vizzuality.atlassian.net/browse/LANDGRIF-1386


client/src/containers/delete-user/component.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const [open, setOpen] = useState(false);
+  const { mutate: deleteUser, isLoading } = useDeleteUser();
+
+  const handleDeleteUser = () => {
+    deleteUser(id, {
+      onSettled: () => setOpen(false),
+      onError: () => toast.error('There was an error deleting the user. Try again later.'),
+      onSuccess: () => toast.success('User deleted successfully'),
+    });
+  };
```
davidsingal, line 24: here we could use `useCallback`

client/src/containers/delete-user/component.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <>
+      <Button size="xs" variant="secondary" danger onClick={() => setOpen(true)}>
+        Delete user
+      </Button>
+      <Modal
+        theme="minimal"
+        title={`Delete user "${userName}"`}
+        open={open}
+        onDismiss={() => setOpen(false)}
```
davidsingal, line 35: here also, `useCallback`

client/src/containers/delete-user/component.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <div>
+            <h1 className="text-base leading-7 font-semibold text-gray-900 mb-1">
+              Delete user &quot;{userName}&quot;
+            </h1>
+            <p className="text-sm text-gray-500 mb-4">
+              All of this user data will be permanently removed from our servers forever. This
+              action cannot be undone.
+            </p>
+            <div className="flex justify-end gap-4">
+              <Button variant="white" onClick={() => setOpen(false)}>
```
davidsingal, line 52: reuse `onDismiss` function you used above

client/src/containers/edit-user/component.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <>
+      <Button size="xs" variant="white" className="!text-sm" onClick={() => setOpenModal(true)}>
+        Edit
+      </Button>
+      <Modal open={open} title={`Edit user ${userName}`} onDismiss={() => setOpenModal(false)}>
+        <UserForm user={user}>
+          <div className="flex justify-between flex-1 mr-2.5 px-0.5">
+            <DeleteUser id={user.id} userName={userName} />
```
davidsingal, line 29:
```suggestion
            <DeleteUser data={user} />
```
I'd prefer to pass the user object, and `userName` should be created inside `DeleteUser` component

client/src/containers/delete-user/component.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Button } from 'components/button';
+import Modal from 'components/modal/component';
+import { useDeleteUser } from 'hooks/users';
+
+type DeleteUserProps = {
+  userName: string;
+  id: string;
+};
+
+const DeleteUser = ({ id, userName }: DeleteUserProps) => {
```
davidsingal, line 14:
```suggestion
const DeleteUserModal = ({ id, userName }: DeleteUserProps) => {
```
I'd prefer to add more info in the component naming.

client/src/containers/edit-user/user-form.tsx
```diff
@@ -0,0 +1,96 @@
...
+    defaultValues: {
+      ...pick(user, 'fname', 'lname', 'displayName', 'email'),
+      roles: user.roles.map((role) => role.name),
+    },
+  });
+
+  const roleOptions = useMemo(
+    () => Object.values(RoleName).map((role) => ({ label: role, value: role })),
+    [],
+  );
```
davidsingal, line 33: `useMemo` is to much for an array of only 2 items, you could remove it

client/src/containers/edit-user/user-form.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  const [roleValue, setRoleValue] = useState<Option<string>[]>(
+    user.roles.map((role) => ({ label: role.name, value: role.name })),
+  );
```
davidsingal, line 37:
```suggestion
  const [roleValue, setRoleValue] = useState<Option<string>[]>(roleOptions);
```
Couldn't be like this?

client/src/types.d.ts
```diff
@@ -362,10 +362,10 @@ export type Task = {
 // User
 export type User = {
   displayName: string;
```
davidsingal, line 364: `displayName` will be removed in the API

davidsingal (changes requested):

client/src/containers/edit-user/component.tsx
barbara-chaves: I created the useName outside because it is used by the parent element too

client/src/containers/edit-user/user-form.tsx
barbara-chaves: It can’t because the  initial state is related to the user data coming from the API, and the roleOptions is a hardcoded value

client/src/types.d.ts
barbara-chaves: I will remove this property and replace it on components by the user full name

client/src/utils/user-full-name.ts
```diff
@@ -0,0 +1,13 @@
+import type { User } from 'types';
+
+const getUserFullName = (user: User, replaceByEmail = false) => {
```
andresgnlez, line 3: Instead of a `replaceByEmail = false` parmeter, I would suggest something like an `options` object. Why? Well, if in the future there are more options to contemplate, with the current approach, you will need to add more parameters to the function, meaning updating the function where is used and forcing to specific all values eventually.

If you have something like 
```
options = {
  replaceByEmail: false,
  useOnlyFirstName: true,
  someRandomFutureOption: false,
},

// your logic
if (options.replaceByEmail) //...
if (options.useOnlyFirstName) //...
if (options.someRandomFutureOption) //...
```

This way you only need to work extending the business logic of the function and forget about updating parameters across the codebase.

TLDR: the code si completely fine, it's just a matter of scalability

andresgnlez: Also, don't forget to type the return of the function with `:string` :)

client/src/utils/user-full-name.ts
```diff
@@ -0,0 +1,13 @@
...
+  if (!user) return '';
+  const { fname, lname, email } = user;
+  if (fname) {
+    return `${fname}${` ${lname}` || ''}`;
```
andresgnlez, line 7: Isn't it simpler just ``${fname} ${lname}`` (notice the space between the two variables) ? Or am I missing something? 🤔

client/src/containers/user-dropdown/component.tsx
```diff
@@ -23,6 +24,8 @@ const UserDropdown: React.FC = () => {
 
   const handleSignOut = useCallback(() => signOut({ callbackUrl: '/auth/signin' }), []);
 
+  const userName = useMemo(() => getUserFullName(user), [user]);
```
andresgnlez, line 27: No need of using `useMemo` here, the operation is not that expensive to optimize it.

client/src/utils/user-full-name.ts
andresgnlez: Also (and I finish, for real), I think is more useful for you, and other developers, knowing the API of your function is `getUserFullName(user, { replaceByEmail: true })` instead of `getUserFullName(user, true)` (as you don't what that `true` really means and would need to check on the function)

andresgnlez:
image
@barbara-chaves  I tried to update my info but the _save_ button is disabled, is this expected? Also, can a user have the roles _admin_ and _user_ at the same time? Sounds weird to me 🤔

client/src/utils/user-full-name.ts
barbara-chaves: Great review! Thanks!

client/src/utils/user-full-name.ts
barbara-chaves: I cant use `${fname} ${lname}` because it can return a 'undefined undefined' for example, but my logic isn't good either. I will improve the function.

andresgnlez: You can check in the `if` if both variables exist, if so, print both names, otherwise, it will print what you have in the final `return` of the function (ideally `fname`)

andresgnlez (approved):

PR #1034, opened 2023-07-19
Client/feature/default unknown suppliers

barbara-chaves:
### General description

This PR adds a default value to the intervention create/edit form for newT1SupplierId and newProducerId

### Testing instructions

When creating or editing an intervention:

- The fields of Supplier (Tier 1 supplier and Producer) should always have a value. 
- The default value should be 'Unknown': when the user didn’t select any option or when the user clears the field 
- The fields should be clearable unless the value is Unknown 

https://github.com/Vizzuality/landgriffon/assets/48164343/c5e0e1f5-90f3-4e07-80d8-fd8a1ae346e5

### Tasks
https://vizzuality.atlassian.net/browse/LANDGRIF-1448
https://vizzuality.atlassian.net/browse/LANDGRIF-1449

### DIFF
--- a/client/CHANGELOG.md
+++ b/client/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 ### Changed
 
 - From update user form: removed 'display name' adn added 'company title' [LANDGRIF-1343](https://vizzuality.atlassian.net/browse/LANDGRIF-1343)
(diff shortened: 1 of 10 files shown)


andresgnlez (approved):

alexeh:
@barbara-chaves I just tested the behaviour in the preview and it works great! 

Some suggestions from my end (_just suggestions_):

1. It would be great if you add this new behaviour to the existing test suite
2. Could you please clean the history removing (dropping, squashing) the revert commits? For the sake of having a cleaner tree :)

PR #1043, opened 2023-07-26
Greenhouse gas emissions from farm production

BielStela:
### General description

Includes in the h3 ingestion makefile the download and conversion to h3 the set of rasters with the ghg per farm production (1:1 with mapspam)

## Checklist before merging


BielStela:
This PR still needs some work to do! 
1. The files stored in the bucket  `landgriffon-raw-data/processed/ghg_emissions/` which unit are they? 
2. Should be 1:1 with mapspam but currently we have extra mapspam aggregations like `wheaocerpotaopul`. How does this fit with the current datasets?
3. The creation scripts should be reviewed, and presented here in such a way that anyone can just run the thing and the raster files will be created easily. The scripts should be place under `data/preprocessing/farm_ghg/` (I already included some files but needs revision @mikeharfoot  
4. The rasters in `landgriffon-raw-data/processed/ghg_emissions/` have notable outliers that shouldn't be there and need to be dealt with or mitigated when the rasters are being created instead of clipping the values before ingesting. for example:   `SUGC_per_t_production` go from 0 to 6729.6 but with 97% of the values between 0 and 1.

mikeharfoot:
I edited the R code to change the projection and this reduces the outliers. Instead of projecting the impacts (e.g. GHGs) to WGS84, then dividing by grid cell production, I'm converting the production to the gall-peters grid and calculating the emission rate in that CRS then reprojecting that to WGS84 - which removes large outliers.

Have committed the revised code to this PR and replace the impact files in the rasters in landgriffon-raw-data/processed/ghg_emissions/ folder.

data/preprocessing/farm_ghg/Reproject_impacts.R
```diff
@@ -0,0 +1,283 @@
...
+new.packages <- wants[!(wants %in% installed.packages()[,"Package"])]
+if(length(new.packages)) install.packages(new.packages)
+lapply(wants, library,character.only = T)
+
+#devtools::install_version("terra","1.7-29")
+
+
+
+#Generalise the data path
+data.path<-"C:/Users/mikeha/Work/Spatial data/Environmental footprint of global food production/"
```
elpamart, line 35: Can we save the raw data in the s3 bucket under the `raw data` folder so we can all have access to it to reproduce the analysis?

elpamart (changes requested):

BielStela:
MOVED to https://github.com/Vizzuality/landgriffon/pull/1065

PR #1074, opened 2023-10-09
Clean h3 and contextuals tables

BielStela:
Adds script util to delete dangling non-linked h3 tables and contextual layers entries.

It removes tables that start with `h3_grid*` and are not in the `h3_data` table. Contextual layer entries that don't have an id listed in h3-data and are not raster tiles (don't have a tile url) are considered dangling and deleted too

### DIFF
--- a/data/h3_data_importer/delete_h3_tables.py
+++ b/data/h3_data_importer/delete_h3_tables.py
@@ -10,8 +10,9 @@ logging.basicConfig(level=logging.INFO)
 
 
 @click.command()
+@click.option("--drop-contextuals", is_flag=True)
 @click.option("--dry-run", is_flag=True)
-def main(dry_run: bool):
+def main(drop_contextuals: bool, dry_run: bool):
     with psycopg.connect(get_connection_info()) as conn:
         with conn.cursor() as cursor:
             # find all the tables that start with h3_grid*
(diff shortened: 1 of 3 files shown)


data/h3_data_importer/delete_h3_tables.py
```diff
@@ -0,0 +1,62 @@
...
+log = logging.getLogger("delete_h3_tables")
+logging.basicConfig(level=logging.INFO)
+
+
+@click.command()
+@click.option("--drop-contextuals", is_flag=True)
+@click.option("--dry-run", is_flag=True)
+def main(drop_contextuals: bool, dry_run: bool):
+    with psycopg.connect(get_connection_info()) as conn:
+        with conn.cursor() as cursor:
```
alexeh, line 17: n00b question: why do we need a different context here? Does the cursor needs to be closed? Could we explicitly do something like ```cursor.close()``` instead of instantiating a new context?

data/h3_data_importer/delete_h3_tables.py
```diff
@@ -0,0 +1,62 @@
...
+
+from utils import get_connection_info
+
+log = logging.getLogger("delete_h3_tables")
+logging.basicConfig(level=logging.INFO)
+
+
+@click.command()
+@click.option("--drop-contextuals", is_flag=True)
+@click.option("--dry-run", is_flag=True)
```
alexeh, line 14: Where is this dry-run set to False, so the code actually cleans the DB?

I am probably missing something, but what is the purpose of those flags in the actual code? Wouldn't be easier just to add some tests? This is a good feature to add tests

data/h3_data_importer/delete_h3_tables.py
```diff
@@ -0,0 +1,62 @@
...
+                """SELECT table_name
+            FROM information_schema.tables as tables
+            LEFT JOIN  h3_data on h3_data."h3tableName" = tables.table_name
+            WHERE tables.table_name LIKE 'h3_grid%' and
+            h3_data."h3tableName" is null;
+            """
+            )
+            tables_to_drop = cursor.fetchall()
+            if tables_to_drop:
+                if not dry_run:
```
alexeh, line 29: Not sure why we need this

data/h3_data_importer/delete_h3_tables.py
```diff
@@ -0,0 +1,62 @@
...
+                cursor.execute(
+                    """SELECT contextual_layer.id
+                FROM contextual_layer
+                LEFT JOIN h3_data ON contextual_layer.id = h3_data."contextualLayerId"
+                WHERE h3_data."contextualLayerId" IS NULL AND contextual_layer."tilerUrl" IS NULL;
+                """
+                )
+                contextuals_to_drop = cursor.fetchall()
+                if contextuals_to_drop:
+                    if dry_run:
```
alexeh, line 47: Same as https://github.com/Vizzuality/landgriffon/pull/1074/files#r1401505553.

alexeh (commented):
Thanks @BielStela, I don't have any objections, but I would like to clarify some questions before pushing this forward

data/h3_data_importer/delete_h3_tables.py
BielStela: For context, `click` is the lib used for cli stuff. When an option is set as a flag it is only set to true when it is present as a argument in the command line. The absence of `--dry-run`  in the _cli_ will make `dry_run` eval to `False` in the code.

data/h3_data_importer/delete_h3_tables.py
BielStela: I'm not sure either xd. Since it is deleting data I wanted to be sure that it is dropping what I want before running this for the real but since this will be mostly called from inside the system it does not make much sense really. I does not hurt tho. I don't know, up to you!

PR #1081, opened 2023-10-16
fixed when to show data table

davidsingal:
Fixed an issue where the table is showing when there is not data and a uploading is in progress.


elpamart:
LGTM!

PR #1100, opened 2023-12-13
Feature/chart legend filters

mbarrenechea:
### General description

This PR add the ability to filter area charts from the legend. It will fade-out the rest of legend items and the chart will change with the new data. In case there aren't children, only the area parent will be shown. We have also took care about the "others" case

### Designs

https://www.figma.com/file/7bmF29CIXDJwCpuloW5Nt6/%F0%9F%94%B5-Landgriffon---Redesign_BLUE?node-id=3121%3A105864&mode=dev

image


davidsingal (approved):
it works really well for me

PR #1113, opened 2024-01-18
 LANDGRIF-1509 Fix small number formatting

BielStela:
Improve table/number readability. Avoid using `m` `µ` abbreviations for values < 1 and uses SI suffixes for > 1000.

~@davidsingal I just applied the formatting to the legend component which I thick was the lowest hanging fruit of them all, if you can please do the same (or the proper way to do this) to the other places where we format would be nice :)~

Updated for all cases.


davidsingal (approved):
Looks really well

PR #1187, opened 2024-05-13
Gadm preprocessing and ingestion

BielStela:
### General description

Small fix of gadm preprocessing and ingestion with and accidental upgrade of deps.

Detail:
- Replaces sql script with python (x100 faster) in preprocessing since we do not need any db ops to generate the intermediate csv
- Fixes the copy from csv in ingestion forcing the input columns to avoid problems with missing columns
- ~Updates h3ronpy to latest version~ it is broken for large geodataframes
- Updates base image gdal to 3.8.5 which ships with python 3.10. It was using 3.3 gdal which used py 3.8 which made lots of dependencies to be obsolete and hard to manage locally. now the requirements can be installed in any up to date environment without a causing a dependency wreck 

### Testing instructions

RUN THE `seed-data`

## Checklist before merging

- [ x] Branch name / PR includes the related Jira ticket Id.
- [ x] All checks in Continuous Integration workflow pass. (lol)
- [ x] Documentation updated (README, CHANGELOG...) (if required)


data/preprocessing/gadm/Makefile
```diff
@@ -1,34 +1,30 @@
...
-WORKDIR=data/tmp
+WORKDIR=data
 
-all: checksum
+all: upload_results
 	make clean-workdir
 
+upload_results: checksum
+	aws s3 cp $(WORKDIR)/geo_region.zip $(AWS_S3_BUCKET_URL)/processed/geo_region/
+	#rm $(WORKDIR)/geo_region.zip $(WORKDIR)/geo_region.csv
```
elpamart, line 11: should we keep this comment here?

elpamart (approved):
I can not check the ingestion but overall looks good to me!

data/preprocessing/gadm/Makefile
BielStela: deleted it because we clean all the data files after finishing automatically anyway

BielStela:
somehow messed with the requirements and still tweaking and testing locally so it works without unexpected crashes

elpamart (approved):

PR #1226, opened 2025-02-10
Chore/landgrif 1602 refactor import data cqrs

jorge-vizz:
### General description

Refactor the `import-data.consumer.ts` file to execute using a CQRS pattern for clarity and performance.

### Testing instructions

No new tests provided, just make sure no regression tests included

## Checklist before merging


api/src/modules/import-data/cqrs/import-data-processing.handler.ts
```diff
@@ -0,0 +1,40 @@
...
+import { Logger } from '@nestjs/common';
+import {
+  IMPORT_DATA_EVENTS,
+  ImportDataEvent,
+} from 'modules/events/import-data-events/import-data.event-handler';
+import { ExcelImportJob } from '../workers/import-data.producer';
+
+export class StartImportProcessingCommand {
+  constructor(public readonly jobData: ExcelImportJob) {}
+}
```
KevSanchez: You could consider this something minor, but I think it might be good idea to pass the needed parameters through the command instead of the whole job object. Mainly to avoid wondering in the future, which parts of the job object are actually needed by the handler, and for consistency with the other two handlers

api/src/modules/import-data/workers/import-data.consumer.ts
```diff
@@ -46,74 +37,28 @@ export class ImportDataConsumer {
     if (this.isJobStalled(err)) {
```
KevSanchez, line 35: the TODO right above this function could be removed, as it was just done ;)

alexeh (approved):
Thanks @jorge-vizz! It looks well, once you address any potential suggestions, feel free to merge it and we can test it in the dev environment

PR #1240, opened 2025-03-06
[LANDGRIF-1626] analysis filters: uses official short name

agnlez:
### General description

### Designs

### Testing instructions

## Checklist before merging


client/src/containers/analysis-visualization/analysis-filters/indicators/index.tsx
```diff
@@ -28,12 +28,12 @@ const IndicatorsFilter = () => {
...
     const categories = Array.from(new Set(data?.map(({ category }) => category).filter(Boolean)));
 
-    const categoryGroups = categories.map((category) => {
+    return categories.map((category) => {
       const indicators = data?.filter((indicator) => indicator.category === category);
       const categoryOptions = indicators.map(
         (indicator) =>
           ({
-            label: indicator.metadata?.short_name,
+            label: indicator.shortName ?? indicator.name,
```
alexeh, line 36: Where is this fallback being triggered? i.e I tested nulloshing the shortName, but in the filters (map, table) does not fallback to the name, leaving the label empty

image

alexeh (approved):
Cheers bro. I left a minor comment and I'd like your opinion.

Not a big issue, as of now we don't enforce the shortName being non-nullable in the model (we should) but in real life this should not happen.

So unless you see the falling back thing is not working well in your side and you'd like to do something about, you can merge this from my side

client/src/containers/analysis-visualization/analysis-filters/indicators/index.tsx
agnlez: For the record: we tested this and the API was returning `""` (empty string) instead of `null`, that's why the fallback did not trigger. We agree there should not be empty values but `null` when the value is not present so the fallback triggers as expected.

PR #1244, opened 2025-03-12
Prototype for GTSWU_NEW indicator

elpamart:
### General description
The PR introduces a notebook for calculating the new Water Gap to Sustainable Water Use (WGSWU_NEW) indicator. Key updates include:

- **New Indicator**: The short code for the indicator is WGSWU_NEW.
- **Baseline Data Ingestion**: Baseline water stress (BWS) data from Aqueduct v4 has been incorporated, including the bws_raw category, which is essential for the calculation. This dataset has been linked to the new short code.
- **New SQL Functions**: The notebook introduces two new functions:
          - `SQL_BWS_IN_STRESSED_AREAS`
          - `SQL_GET_STRESSED_AREA_PORTION`
         
These functions are integrated into the overall query to compute **BWS in stressed areas** and the **stressed area portion**.

- **Calculation Process**: The notebook outlines the formula for deriving GTSWU_NEW, which involves calculating the excess of BWS, multiplying it by the stressed area portion and water withdrawals


fgassert:
For the stressed_area_portion query it shouldn’t be multiplying by BWS, just taking the sum of production where BWS>0.4

instead of
`SELECT sum(h3prod.%I * (CASE WHEN h3ind.%I > 0.4 AND h3ind.%I< 9999 THEN h3ind.%I ELSE 0 END))/sum(h3prod.%I)`

it could be
`SELECT sum(h3prod.%I * (CASE WHEN h3ind.%I > 0.4 AND h3ind.%I< 9999 THEN 1 ELSE 0 END))/sum(h3prod.%I)`

or simply
`SELECT sum(CASE WHEN h3ind.%I > 0.4 AND h3ind.%I< 9999 THEN h3prod.%I ELSE 0 END)/sum(h3prod.%I)`

